PR #8, opened 2021-08-03
Add depends_on to examples and docs

futuretea:
1. Add depends_on to examples and docs
2.  fix CRLF to LF

**Related issues**
harvester/harvester#1054


guangbochen (approved):
LGTM

PR #15, opened 2021-09-17
Fix volume delete

futuretea:

**Related issues**
harvester/harvester#1227


guangbochen (approved):
LGTM

PR #23, opened 2022-01-20
Add validate method and check ssh keys in cloud-init

futuretea:

**Related issue**
https://github.com/harvester/harvester/issues/1807


guangbochen (approved):
LGTM

PR #32, opened 2022-03-23
Clusternetwork import Enhancement

futuretea:
**Related issues**
https://github.com/harvester/harvester/issues/2052

### DIFF
--- a/docs/resources/clusternetwork.md
+++ b/docs/resources/clusternetwork.md
@@ -19,7 +19,7 @@ resource "harvester_clusternetwork" "vlan" {
   }
   name                 = "vlan"
   enable               = true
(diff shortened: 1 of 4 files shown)


guangbochen:
does it mean we only support importing the `clusterNetwork` resources instead of allowing them to create it via the TF?

guangbochen:
does it mean we only support importing the `clusterNetwork` resources instead of allowing them to create it via the TF?

futuretea:
> does it mean we only support importing the `clusterNetwork` resources instead of allowing them to create it via the TF?

yes

guangbochen (approved):
LGTM

PR #36, opened 2022-05-06
Add run strategy, fix ssh key name , fix memory

futuretea:
- Use `run_strategy` instead of `start`
- Read vm memory from limit
- optimize SSH key name display

**Related issues**
https://github.com/harvester/harvester/issues/2225
https://github.com/harvester/harvester/issues/2222


docs/resources/virtualmachine.md
```diff
@@ -64,12 +64,12 @@ resource "harvester_virtualmachine" "ubuntu20-dev" {
   cpu    = 2
   memory = "2Gi"
 
-  start        = true
+  run_strategy = "Always"
```
guangbochen, line 67: the default run strategy in Harvester is `RerunOnFailure`

examples/resources/harvester_virtualmachine/resource.tf
```diff
@@ -49,12 +49,12 @@ resource "harvester_virtualmachine" "ubuntu20-dev" {
   cpu    = 2
   memory = "2Gi"
 
-  start        = true
+  run_strategy = "Always"
```
guangbochen, line 52: ditto

internal/provider/virtualmachine/schema_virtualmachine.go
```diff
@@ -19,10 +23,28 @@ func Schema() map[string]*schema.Schema {
...
 		constants.FieldVirtualMachineStart: {
 			Type:     schema.TypeBool,
 			Optional: true,
-			Default:  true,
+			Deprecated: fmt.Sprintf(`
+please use %s instead of this deprecated field:
+	%s = true  ==>  %s = "%s"
+	%s = false  ==>  %s = "%s"
+`, constants.FieldVirtualMachineRunStrategy,
+				constants.FieldVirtualMachineStart, constants.FieldVirtualMachineRunStrategy, kubevirtv1.RunStrategyAlways,
```
guangbochen, line 46: ditto

docs/resources/virtualmachine.md
futuretea: If use RerunOnFailure，user can't start/stop vm by terraform

guangbochen: both `Always` and `RerunOnFailure` will have the same behavior of start and stop, are we using sub-resource action to manage it?
image

futuretea: > are we using sub-resource action to manage it?
No, RerunOnFailure  is ok. I thought RerunOnFailure's behavior is like Manual before.

Changed to RerunOnFailure

examples/resources/harvester_virtualmachine/resource.tf
futuretea: @guangbochen updated

guangbochen (approved):
LGTM

PR #38, opened 2022-06-21
Change route_gateway and route_cidr to computed

futuretea:
Change route_gateway and route_cidr to computed to ignore route_gateway and route_cidr changes
**Related issues**
https://github.com/harvester/harvester/issues/2261


internal/provider/network/schema_network.go
```diff
@@ -39,12 +39,14 @@ func Schema() map[string]*schema.Schema {
 		constants.FieldNetworkRouteCIDR: {
 			Type:          schema.TypeString,
 			Optional:      true,
+			Computed:      true,
```
guangbochen, line 42: IIRC, both CIDR and gateway can be configured by the users, so computed may not be a valid solution.
https://www.terraform.io/plugin/sdkv2/schemas/schema-behaviors#computed

futuretea: Users can set value to a Computed field

guangbochen: Does it mean the `computed` definition can help to merge the output after `tf apply` and it is also configurable if needed?

futuretea: Yes

guangbochen (approved):
LGTM, thanks.

PR #40, opened 2022-07-07
should report not found error when data source read

futuretea:
For data source, if the resource cannot be found in the cluster based on the name and namespace, an error message should be reported.

otherwise, the user might use an empty id in other resources

**Related issues**
https://github.com/harvester/harvester/issues/2461

**Test Plan**
1. create a network `vlan1` in the namespace `harvester-public` form UI side
2. create a data source harvester_network.vlan1, do not specify namespace, It will default to `default`
```hcl
data "harvester_network" "vlan1" {
  name      = "vlan1"
}
```
3. terraform apply
4. users will get an error
```
Error: network-attachment-definitions.k8s.cni.cncf.io "vlan1" not found
```
6. add correct namespace to the data source
```hcl
data "harvester_network" "vlan1" {
  name      = "vlan1"
}
```
7. terraform apply will be success


chrisho (approved):
LGTM

PR #48, opened 2022-08-10
Resolving vm running state on creation

matttrach:
Signed-off-by: matttrach <matttrach@gmail.com>

Without this change the waiting on state function never returns, and the provider eventually times out.
I think the acceptance test bypasses this by validating that the server exists in harvester, but not that the state is in an expected target state.


matttrach:
this was not the correct solution

TSmike17:
Currently suffering from this issue, was it ever resolved?

PR #53, opened 2022-09-27
Fix pvc.StorageClassName order

vulkoingim:
If both `storage_class_name` and `image` are provided, then `storage_class_name` gets overwritten by the one derived from the image.

This changes the order of assignment for the inline disk spec and adds a check for the volume resource, so if `storage_class_name` is provided, it doesn't get overwritten.

Tested it by building the provider locally and verifying that it creates the PVC with the correct `StorageClass`, both with specifying it and without.


futuretea:
Since The storageclass of image volume cannot be modified,
so the `storage_class_name` field in `harvester_image` is a `Computed` field in v0.5.2
will rename the field `storage_class_name` to `volume_storage_class_name` in https://github.com/harvester/terraform-provider-harvester/pull/54
and add a new field `storage_class_name` as a template to help user to specify storageClass parameters
more info: https://github.com/futuretea/harvester/blob/storage-tiering/enhancements/20220531-storage-tiering.md

PR #57, opened 2022-10-21
Update image volume examples and docs

futuretea:
use harvester_image id instead of image namspace/name

to make sure that the dependencies between resources are correct


chrisho (approved):
LGTM

PR #65, opened 2022-12-12
Bump version to 0.6.1

futuretea:

bump version to v0.6.1


masteryyh (approved):
Tested with creating some networks, images and virtual machines and works like a charm, thanks.

PR #60, opened 2022-11-17
Fix network apply error with dhcpserver

futuretea:
Signed-off-by: futuretea <Hang.Yu@suse.com>

**Related issues**
https://github.com/harvester/harvester/issues/2857

**Test plan**

- Setup test env by https://github.com/harvester/terraform-provider-harvester/wiki/PR-Test-Env-Setup for `Reviwer`
- Setup test env by https://github.com/harvester/terraform-provider-harvester/wiki/Use-the-master-head-docker-image-for-testing for `Tester`
- Create a test config file
```bash
vim main.tf
```
**Network**
- Add resource `harvester_network` with `route_mode = "auto"` and `route_dhcp_server_ip = "<replace to your dhcp server>"`
```hcl
resource "harvester_network" "mgmt-vlan1" {
  name      = "mgmt-vlan1"
  namespace = "harvester-public"

  vlan_id = 1

  route_mode           = "auto"
  route_dhcp_server_ip = "192.168.5.1"

  cluster_network_name = "mgmt"
}
```
- apply 
```bash
terraform apply -auto-approve
```
- apply again
```bash
terraform apply -auto-approve
```

PR #68, opened 2022-12-14
Fix update network description

futuretea:

Fixe update network description did not take effect found in https://github.com/harvester/terraform-provider-harvester/pull/60


chrisho (approved):
Verified and LGTM, thanks

PR #73, opened 2023-02-02
Update list of OS and ARCH to target

jpvlsmv:
The current terraform-provider-scaffolding repository offers builds for Freebsd, Windows, Linux, and Darwin.

In my particular case, I'm running on Windows, and currently get an error message of 
> Provider registry.terraform.io/harvester/harvester v0.6.1 does not have a package available for your current platform, windows_amd64.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -16,11 +16,18 @@ builds:
   ldflags:
     - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
   goos:
+    - freebsd
+    - windows
     - linux
     - darwin
   goarch:
     - amd64
+    - '386'
+    - arm
     - arm64
+  ignore:
+    - goos: darwin
+      goarch: '386'
   binary: '{{ .ProjectName }}_v{{ .Version }}'
 archives:
 - format: zip


.goreleaser.yml
```diff
@@ -16,11 +16,18 @@ builds:
   ldflags:
     - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
   goos:
+    - freebsd
+    - windows
```
futuretea, line 20: 5351aa7e85011caa163f670450d6a08
e312a1b9b1e7644a3cfebf9dc87c79d
It does not compile successfully on freebsd and windows

jpvlsmv:
As long as github.com/c9s/goprocinfo is a dependency (it is Linux-only) this module can not run on any other platforms :(

It's not a direct dependency.
terraform-provider-harvester depends on github.com/harvester/harvester
github.com/harvester/harvester depends on kubevirt.io/kubevirt
kubevirt.io/kubevirt depends on c9s/goprocinfo in its pkg/monitoring/domainstats/downwardmetrics/hostmetrics.go

It "should" be possible to stub out the dependency (assuming that I don't need to monitor the host metrics when using the provider) but I'm not enough of a golang programmer to fully understand how it would be done.

PR #78, opened 2023-03-30
Add bootstrap

FrankYang0529:
issue: https://github.com/harvester/harvester/issues/1797

**Test plan**

1. Create a new Harvester cluster and don't login.
2. Build terraform-harvester-provider.
```
CGO_ENABLED=0 go build -mod=vendor -o bin/terraform-provider-harvester .
mkdir -p ~/.terraform.d/plugins/registry.terraform.io/harvester/harvester/0.0.0-master/linux_amd64
cp bin/terraform-provider-harvester ~/.terraform.d/plugins/registry.terraform.io/harvester/harvester/0.0.0-master/linux_amd64/terraform-provider-harvester_v0.0.0-master
```
3. Create a temporary folder and put the following content in `main.tf` (remember to change `api_url).
```
terraform {
  required_providers {
    harvester = {
      source = "harvester/harvester"
      version = "0.0.0-master"
    }
  }
}

provider "harvester" {
  alias = "bootstrap"

  api_url    = "https://192.168.3.131"
  bootstrap = true
}

resource "harvester_bootstrap" "admin" {
  provider = "harvester.bootstrap"

  initial_password = "admin"
  password = "passwdpasswd"
}

provider "harvester" {
  alias = "admin"

  api_url    = "https://192.168.3.131"
  kubeconfig = harvester_bootstrap.admin.kubeconfig
}

resource "harvester_volume" "test-disk" {
  provider = "harvester.admin"

  name      = "test-disk"
  namespace = "default"

  size = "1Gi"
}

resource "harvester_image" "ubuntu20" {
  provider = "harvester.admin"

  name      = "ubuntu20"

  display_name = "ubuntu-20.04-server-cloudimg-amd64.img"
  source_type  = "download"
  url          = "http://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64.img"
}

resource "harvester_virtualmachine" "ubuntu20" {
  provider = "harvester.admin"

  name                 = "ubuntu20"
  namespace            = "default"
  restart_after_update = true

  description = "test ubuntu20 raw image"
  tags = {
    ssh-user = "ubuntu"
  }

  cpu    = 2
  memory = "2Gi"

  efi         = true
  secure_boot = true

  hostname        = "ubuntu20"
  reserved_memory = "100Mi"
  machine_type    = "q35"

  network_interface {
    name           = "nic-1"
    wait_for_lease = true
  }

  disk {
    name       = "rootdisk"
    type       = "disk"
    size       = "10Gi"
    bus        = "virtio"
    boot_order = 1

    image       = harvester_image.ubuntu20.id
    auto_delete = true
  }

  cloudinit {
    user_data    = <<-EOF
      #cloud-config
      password: 123456
      chpasswd:
        expire: false
      ssh_pwauth: true
      package_update: true
      packages:
        - qemu-guest-agent
      runcmd:
        - - systemctl
          - enable
          - '--now'
          - qemu-guest-agent
      EOF
    network_data = ""
  }
}
```
4. Run `terraform init` and `terraform apply`. Terraform should update password to `passwdpasswd` and create a new image and vm.


futuretea:
Refer to https://github.com/harvester/harvester/pull/3673, It seems like we won't disable MCM in 1.2.0. So users can use terraform-provider-rancher2 to do this.

https://github.com/harvester/harvester/issues/1797#issuecomment-1422869575

futuretea (commented):
Refer to https://github.com/harvester/harvester/pull/3673, It seems like we won't disable MCM in 1.2.0. So users can use terraform-provider-rancher2 to do this.

https://github.com/harvester/harvester/issues/1797#issuecomment-1422869575

belgaied2:
This PR is interesting because I am encountering issues where it is not possible to use the Harvester provider where the KUBECONFIG file does not yet exist. I get the following error during `terraform plan` phase:
`Error: the server could not find the requested resource (get settings.harvesterhci.io server-version)` 

Is this solved somewhere?  Can the `plan` phase be successful without an existing kubeconfig file ?

FrankYang0529:
Hi @belgaied2, thanks for the testing. I updated the code. Now, it can handle `terraform plan` without an existing kubeconfig file. Thanks.

internal/provider/bootstrap/schema_bootstrap.go
```diff
@@ -0,0 +1,43 @@
...
+		constants.FieldBootstrapInitialPassword: {
+			Type:         schema.TypeString,
+			Optional:     true,
+			Sensitive:    true,
+			ForceNew:     true,
+			Default:      "admin",
+			ValidateFunc: validation.NoZeroValues,
+			Description:  "Default password in the harvester",
+		},
+		constants.FieldBootstrapPassword: {
```
m-ildefons, line 21: Same as with the kubeconfig in base64 format, there should be a warning in the documentation, informing the user that this sensitive information will be persisted in clear text in the Terraform state file

internal/config/config.go
```diff
@@ -0,0 +1,40 @@
...
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/harvester/terraform-provider-harvester/pkg/client"
+)
+
+type Config struct {
+	Bootstrap   bool
+	APIURL      string
```
m-ildefons, line 15: Personally, I would make this part of the bootstrap resource.
Thereby the `Config` structure can not contain contradicting information between this field and the API url that is encoded in the kubeconfig in the field below.
Since the API url is ignored in non-bootstrap mode, it create a confusing situation.

internal/provider/volume/resource_volume.go
```diff
@@ -131,7 +143,10 @@ func resourceVolumeImport(d *schema.ResourceData, obj *corev1.PersistentVolumeCl
 
 func resourceVolumeRefresh(ctx context.Context, d *schema.ResourceData, meta interface{}) resource.StateRefreshFunc {
 	return func() (interface{}, string, error) {
-		c := meta.(*client.Client)
+		c, err := meta.(*config.Config).K8sClient()
```
m-ildefons, line 146: This now constructs a new client instance every time this function runs.
Keep in mind, this function will be run in a loop to watch the state of a volume (e.g. while it is being removed).
Instead the client should be handed into this function from outside, so it can be keep throughout multiple invocations.

m-ildefons (changes requested):
Hey PoAn,
this PR is cool and implements a very useful functionality. However I think this PR needs a bit more work:

- The provider configuration can be simplified by moving the API url property to the bootstrap resource, which is the only place where it is utilized
- There is a problem with the way the new `Config` object works, because now a new client is instantiated at the beginning of every operation. This can lead to problems, because the Kubernetes API is cached and may return stale data. This isn't a problem when re-using the same client instance, because the client keeps a cache too.
- I disapprove of the lack of tests for the new resource
- I disapprove of the lack of documentation, particularly when there are confidential information being persisted in clear text in the Terraform state file.

internal/config/config.go
FrankYang0529: Moved it to bootstrap resource.

internal/provider/bootstrap/schema_bootstrap.go
FrankYang0529: How do you check terraform state? It shows sensitive data on command:

```
> terraform state show 'harvester_bootstrap.admin'
# harvester_bootstrap.admin:
resource "harvester_bootstrap" "admin" {
    api_url                = "https://192.168.3.131"
    id                     = "token-kjh9w"
    initial_password       = (sensitive value)
    kubeconfig             = "~/.kube/config"
    password               = (sensitive value)
    should_update_password = false
}
```

internal/provider/volume/resource_volume.go
FrankYang0529: Changed to reuse `K8sClient`.

internal/provider/network/resource_network.go
```diff
@@ -55,7 +58,10 @@ func resourceNetworkCreate(ctx context.Context, d *schema.ResourceData, meta int
 }
 
 func resourceNetworkUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	c := meta.(*client.Client)
+	c, err := meta.(*config.Config).K8sClient()
```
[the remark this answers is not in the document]
FrankYang0529: The result from API and cache will be eventually synced. Not sure why you think it's an issue. But anyway, we change to reuse the client.

internal/provider/bootstrap/schema_bootstrap.go
m-ildefons: Just look in the the `.tfstate` or `.tfstate.backup` files. The content of resources will be persisted there, even if they are marked as sensitive values:
https://developer.hashicorp.com/terraform/tutorials/configuration-language/sensitive-variables#sensitive-values-in-state

I don't see that as a huge problem, but users need to be made aware that this is happening.

m-ildefons (approved):
LGTM

PR #82, opened 2023-08-22
Allow -1 value for bond_miimon uplink bond option

tbregolin:
While not a valid option per the Kernel's documentation, Harvester uses -1 as a special value to preserve whatever the current value is. If -1 is not allowed and the option is simply omitted for the vlanconfig, the Terraform provider will always show a change of the state from -1 to "null" at every run.

The vlanconfig cannot be changed without stopping all the VM instances using it, which is why it would be useful to allow the special value that preserves the existing configuration.

### DIFF
--- a/internal/provider/vlanconfig/schema_vlanconfig_uplink.go
+++ b/internal/provider/vlanconfig/schema_vlanconfig_uplink.go
@@ -34,7 +34,7 @@ func resourceUplinkSchema() map[string]*schema.Schema {
 		constants.FieldUplinkBondMiimon: {
 			Type:         schema.TypeInt,
 			Optional:     true,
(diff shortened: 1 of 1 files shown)


huberpoa (approved):

mauriiciodias (approved):

FrankYang0529 (approved):
As `Miimon` in kubebuiler validation, yes, we should allow at least `-1`.

https://github.com/harvester/network-controller-harvester/blob/3551e5604377ede8c119cf8a3c8af063d25cd1bc/pkg/apis/network.harvesterhci.io/v1beta1/vlanconfig.go#L56-L59

LGTM. Thanks for the contribution!

Yu-Jack (approved):

PR #84, opened 2023-10-16
bump golang to v1.20.10

FrankYang0529:
**Problem:**
We used go 1.18, which is EOL in 2023/02.

**Solution:**
Bump golang to v1.20.10.

**Related Issue:**
https://github.com/harvester/security/issues/38


go.mod
```diff
@@ -59,32 +59,35 @@ replace (
 replace github.com/harvester/harvester => github.com/harvester/harvester v0.0.2-0.20230209065814-98f56beae339
 
 require (
-	github.com/harvester/harvester v1.1.1
-	github.com/harvester/harvester-network-controller v0.3.1
+	github.com/harvester/harvester v1.1.2-rc8
```
Yu-Jack, line 62: Is that okay to use rc here?

FrankYang0529: Yeah, it's good to update to v1.2.0. Thanks.

bk201 (approved):
LGTM!

PR #87, opened 2023-12-22
ci: fix go version with quote for release

FrankYang0529:
Without quote, CI think 1.20 as 1.2.

Ref: https://github.com/harvester/terraform-provider-harvester/actions/runs/7296020228/job/19883228070

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.20
+          go-version: '1.20'
       -
         name: Import GPG key
         id: import_gpg


Yu-Jack (approved):
LGTM, interesting

bk201 (approved):

Yu-Jack:
Add context from https://github.com/actions/setup-go:

> Note: Due to the peculiarities of YAML parsing, it is recommended to wrap the version in single quotation marks:
>
>   go-version: '1.20'
>
> The recommendation is based on the YAML parser's behavior, which interprets non-wrapped values as numbers and, in the case of version 1.20, trims it down to 1.2, which may not be very obvious. Matching an unstable pre-release:

PR #91, opened 2024-04-09
MINOR: Fix go version in readme

brandboat:
now we use go 1.21 instead of v1.20


FrankYang0529 (approved):
LGTM. Thanks for the fix.

bk201 (approved):

PR #92, opened 2024-04-11
ci: change credential path from rancher to harvester

FrankYang0529:
https://github.com/harvester/harvester/issues/5327


Anarkis (approved):
LGTM

PR #95, opened 2024-04-18
Separate push_image steps to another job

FrankYang0529:
https://github.com/harvester/harvester/issues/5327


bk201 (approved):
lgtm

.github/workflows/build.yml
```diff
@@ -9,46 +9,62 @@ on:
...
-    # For multi-platform support
-    - name: Set up QEMU
-      uses: docker/setup-qemu-action@v3
-    - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@v3
+    - name: Download binaries
+      uses: actions/download-artifact@v4
+      with:
+        name: binaries_artifact
+        path: ./bin/
```
bk201, line 44: Can you help double-check if the binaries are executable? It seems to be packed with zip.

FrankYang0529: Yeah, we miss it. I add `chmod +x` before copying files to `package` folder. Thank you.

Yu-Jack (approved):

PR #97, opened 2024-05-24
Add timeout blocks for Terraform resources

m-ildefons:
By convention (https://developer.hashicorp.com/terraform/language/resources/syntax#operation-timeouts) terraform resources can have timeouts associated with their operations through `timeout` blocks.
This change sets the default timeout to 2 minutes, except where other defaults were hard-coded before and allows users to specify timeout blocks without causing parsing errors.

Use the contexts to setup and propagate deadlines for each operation according to what is set in the terraform file. Since each API call happens within the same context, the same deadline applies, giving each operation (create, update, read, delete) as a whole the configured timeout rather than setting this timeout on individual API calls of the operation.

Where applicable, the timeout of resources is measured against the observed changes of the API resources or events rather than completion of API calls. This is particularly important for resources where operations can take a lot longer than the API call. E.g. for the Delete operation on a Volume, the time until a matching event is observed is counted. For the Update operation of a VirtualMachine, the time (if necessary) until the VM resource has been restarted and is observed in running state is counted. For a contrast, the Delete operation on a storage class happens almost immediately and the timeout is only counted against the API call.

As an example how timeout blocks can be used, here is an example terraform script for loading a VM image into Harvester from a slow server. Normally this operation should succeed without explicitly setting a timeout, but due to the slow downloadspeed from the server onto the Harvester node, the timeout for creating or updating the resource needs to be significantly increased.

```
terraform {
  required_providers {
    harvester = {
[...]
    }
  }
}

provider "harvester" {
  kubeconfig = "/home/user/.kube.harvester/config.yaml"
}

resource "harvester_image" "opensuse154" {
  name      = "opensuse154"

  display_name = "openSUSE-Leap-15.4.x86_64-NoCloud.qcow2"
  source_type  = "download"
  url          = "https://mirrors.superslow.org/images/openSUSE-Leap-15.4.x86_64-NoCloud.qcow2"

  timeouts {
    create = "20m"
    update = "20m"
  }
}
```

fixes: https://github.com/harvester/harvester/issues/5632
related issues:
- https://github.com/harvester/harvester/issues/5632
- https://github.com/harvester/harvester/issues/5786

### DIFF
--- a/internal/provider/cloudinitsecret/resource_cloudinitsecret.go
+++ b/internal/provider/cloudinitsecret/resource_cloudinitsecret.go
@@ -3,12 +3,10 @@ package cloudinitsecret
 import (
 	"context"
 	"encoding/base64"
-	"errors"
 	"fmt"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -111,18 +109,6 @@ func resourceCloudInitSecretDelete(ctx context.Context, d *schema.ResourceData, 
 		return diag.FromErr(err)
 	}
 
-	stateConf := &resource.StateChangeConf{
-		Pending:    []string{constants.StateImageTerminating, constants.StateCommonActive},
-		Target:     []string{constants.StateCommonRemoved},
-		Refresh:    resourceCloudInitSecretRefresh(ctx, d, meta),
-		Timeout:    d.Timeout(schema.TimeoutDelete),
-		Delay:      1 * time.Second,
-		MinTimeout: 3 * time.Second,
-	}
-	_, err = stateConf.WaitForStateContext(ctx)
-	if err != nil {
-		return diag.FromErr(err)
-	}
 	d.SetId("")
 	return nil
 }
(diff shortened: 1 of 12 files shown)


internal/timeouts/timeouts.go
```diff
@@ -0,0 +1,58 @@
...
+		return context.WithCancel(ctx)
+	}
+
+	// Enforce a minimum timeout to prevent users from setting a ridiculously low
+	// timeout, which could cause Terraform to become unable to modify resources.
+	if minimumTimeout <= d.Timeout(key) {
+		timeout = d.Timeout(key)
+	} else {
+		timeout = minimumTimeout
+	}
```
brandboat, line 55: This seems not work. I have tried set timeout to 1s to create vm image and the execution after 1s not 30s.

m-ildefons: You're right. This flooring of the timeout never worked and my testing wasn't proper.

I've also realized a few points about the SDK and contexts:

- The context for the `CreateContext`, `ReadContext`, etc. functions of the resources is already populated with the respective timeouts and the default is 20 minutes in the SDK. I've overridden the default to be 2 minutes if no explicit timeout is given. This might be a bit short for some resources, but I think 20 minutes is quite excessive.
This also means quite a bit of code is completely unnecessary as long as the context that is passed into the functions is properly propagated to API calls.

- It's impossible create a derived context with a longer timeout than the parent context, but the context also contains other information, so we can't just instantiate a new context. Therefore it's impossible to modify a given timeout, which is why the minimum timeout never worked.
While this means we can't enforce a minimum timeout, this isn't really going to be a common problem anyways.

brandboat: > While this means we can't enforce a minimum timeout, this isn't really going to be a common problem anyways.

Indeed, now we cant set the default timeout to 2 mins, so this should be fine.

brandboat:
Could you also modify the pr description and commit message, the behavior is mismatch to those descriptions now. Many thanks 😃 .

internal/provider/cloudinitsecret/resource_cloudinitsecret.go
```diff
@@ -100,6 +110,19 @@ func resourceCloudInitSecretDelete(ctx context.Context, d *schema.ResourceData,
 	if err != nil && !apierrors.IsNotFound(err) {
 		return diag.FromErr(err)
 	}
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{constants.StateImageTerminating, constants.StateCommonActive},
```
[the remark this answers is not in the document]
m-ildefons: You're right. This whole block of code doesn't make a lot of sense since the cloudinit secret doesn't have a state. It either exists or it doesn't but there is not any intermediate state as there would be with a Pod or similar. Therefore we don't need to wait here, the `DELETE` request should be enough.

internal/provider/cloudinitsecret/resource_cloudinitsecret.go
```diff
@@ -129,3 +152,30 @@ func resourceCloudInitSecretImport(d *schema.ResourceData, obj *corev1.Secret) e
...
+				return obj, constants.StateCommonRemoved, nil
+			}
+			return obj, constants.StateCommonError, err
+		}
+
+		if err = resourceCloudInitSecretImport(d, obj); err != nil {
+			return obj, constants.StateCommonError, err
+		}
+
+		state := d.Get(constants.FieldCommonState).(string)
```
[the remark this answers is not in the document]
m-ildefons: See above. This whole block of code is not sensible.

m-ildefons:
I've changed the PR description to match the behavior of the PR, the commit message and fixed the cloud-init secret resource.

brandboat (approved):
LGTM !

FrankYang0529 (approved):
LGTM. Thanks for the PR.

rvdh:
@FrankYang0529 Can you make a new release containing these improvements? They would help us a lot :)

Thanks 🙏

FrankYang0529:
Yes, we just released v0.6.5. Thanks.

https://github.com/harvester/terraform-provider-harvester/releases/tag/v0.6.5

PR #99, opened 2024-07-06
doc: Add Example Of Network Data

irishgordo:
* adding example of network data

Resolves: doc/add-net-data

### DIFF
--- a/docs/resources/cloudinit_secret.md
+++ b/docs/resources/cloudinit_secret.md
@@ -14,7 +14,7 @@ description: |-
 
 ```terraform
 resource "harvester_cloudinit_secret" "cloud-config-opensuse154" {
-  name      = "cloud-config-opensuse154"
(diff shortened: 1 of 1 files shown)


cclose:
I'm pretty sure that the docs are generated using [`tfplugindocs`](https://github.com/hashicorp/terraform-plugin-docs)

I think you need to update the corresponding file in [`examples/resources/harvester_cloudinit_secret/resource.tf`](https://github.com/harvester/terraform-provider-harvester/blob/45fde4b3f1e0f441f4f347c9fabde756632d259c/examples/resources/harvester_cloudinit_secret/resource.tf#L27)

Not a contributor/maintainer myself, but I was also looking into a documentation issue.

irishgordo:
@cclose - thanks for that mention :smile: :+1:  - I will be sure to update this draft PR with those changes

PR #103, opened 2024-08-26
Tests: Enable Acceptance Tests

m-ildefons:
Enable acceptance tests and fix existing tests and providers

1) Enable running acceptance tests on `make test` command when the user
   places a kubeconfig file into the repository. This kubeconfig is
   necessary since the acceptance tests are checked against a real
   Harvester cluster. If the kubeconfig file is missing, only unit tests
   are run, which doesn't require a Harvester cluster.
2) Fix existing tests. Some existing acceptance tests where using
   invalid terraform resources. These are fixed
3) Fix volume and network resource providers. Running acceptance tests
   revealed bugs in the volume and network resource providers. Both were
   using an incorrect method to determine if a resource had been deleted
   in Harvester. The bug was that the resource providers were waiting
   for a deletion event, despite the fact that the respective
   controllers don't emit such an event. The fix is to observe the
   resource through a periodic read.
4) Fix propagating the cluster network name of a network resource into
   the terraform state. The terraform state needs to keep track of this
   property to determine if the resource definition has changed.
   Otherwise the resource will always count as modified, which is in
   some cases wrong. This bug was also revealed by the acceptance tests.
   
Related-To: https://github.com/harvester/harvester/issues/6415

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -4,7 +4,7 @@ ARG DAPPER_HOST_ARCH
 ENV ARCH $DAPPER_HOST_ARCH
 
 RUN zypper -n rm container-suseconnect && \
-    zypper -n install curl docker gzip tar wget awk zip
+    zypper -n install curl docker gzip tar wget awk zip unzip
 
 # install goimports
 RUN GO111MODULE=on go install golang.org/x/tools/cmd/goimports@v0.1.11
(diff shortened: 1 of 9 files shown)


FrankYang0529 (approved):
LGTM. Thanks for the great fix.

bk201 (approved):

PR #104, opened 2024-09-18
fix: release pipeline

FrankYang0529:
* change actions/checkout to v4
* change actions/setup-go to v5
* change go-version to v1.22
* change goreleaser/goreleaser-action to v6
* add "version: 2" to .goreleaser.yml
* remove darwin from .goreleaser.yml, because we import
longhorn/go-common-libs which uses unix.CLONE_NEWNET

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,12 +1,12 @@
 # This GitHub action can publish assets for release when a tag is created.
 # Currently its setup to run on any tag that matches the pattern "v*" (ie. v0.1.0).
 #
-# This uses an action (hashicorp/ghaction-import-gpg) that assumes you set your 
+# This uses an action (hashicorp/ghaction-import-gpg) that assumes you set your
 # private key in the `GPG_PRIVATE_KEY` secret and passphrase in the `PASSPHRASE`
 # secret. If you would rather own your own GPG handling, please fork this action
 # or use an alternative one for key handling.
 #
-# You will need to pass the `--batch` flag to `gpg` in your signing step 
+# You will need to pass the `--batch` flag to `gpg` in your signing step
 # in `goreleaser` to indicate this is being used in a non-interactive mode.
 #
 name: release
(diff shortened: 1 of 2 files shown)


Yu-Jack (approved):
LGTM, thanks!

bk201 (approved):

rvdh:
I think this change cause the provider to be not available for me on macos:

│ Provider registry.terraform.io/harvester/harvester v0.6.5 does not have a package available for your current platform, darwin_amd64.

:(

bk201:
@FrankYang0529 Can we see if we can enable darwin platform again?

FrankYang0529:
Yeah, @Yu-Jack and I are checking on it.

Yu-Jack:
@FrankYang0529 I think we need to modify longhorn/go-common-libs by creating different system go files. Here are two examples you could try that.

### Example

This example works with different builds.

- `GOOS=linux go build .`
- `GOOS=darwin go build .`

```go
// main.go
package main

import "fmt"

func main() {
	a := Namespace("test")
	fmt.Println(a.Flag())
}
```
```go
// namespace_linux.go
// go:build linux

package main

import (
	"fmt"

	"golang.org/x/sys/unix"
)

type Namespace string

const (
	NamespaceNet = Namespace("net")
)

func (ns Namespace) Flag() uintptr {
	fmt.Println("linux")
	switch ns {
	case NamespaceNet:
		return unix.CLONE_NEWNET
	default:
		return 0
	}
}
```
```go
// namespace_darwin.go
// go:build darwin

package main

import (
	"fmt"
)

type Namespace string

const (
	NamespaceNet = Namespace("net")
)

func (ns Namespace) Flag() uintptr {
	fmt.Println("darwin")
	return 0
}

```

### Example 2

This is a wrong example. It complains the error which is `undefined: unix.CLONE_NEWNET` same as we have.

```go
// namespace_darwin.go
// go:build darwin

package main

import (
	"fmt"

	"golang.org/x/sys/unix"
)

type Namespace string

const (
	NamespaceNet = Namespace("net")
)

func (ns Namespace) Flag() uintptr {
	fmt.Print(unix.CLONE_NEWNET)
	fmt.Println("darwin")
	return 0
}
```

So, we need to add `namespace_linux.go` and `namespace_darwin.go` for longhorn/go-common-libs.

moio:
> @FrankYang0529 @Yu-Jack I think this change cause the provider to be not available for me on macos:
> 
> │ Provider registry.terraform.io/harvester/harvester v0.6.5 does not have a package available for your current platform, darwin_amd64.
> 
> :(

https://github.com/harvester/harvester/issues/6655 tracks that

PR #106, opened 2024-09-27
feat: add Darwin supporting back

Yu-Jack:
## Related Issues

- https://github.com/harvester/harvester/issues/6655
- https://github.com/longhorn/longhorn/issues/9487

## Description

Since go-common-libs did a breaking change:

From:
```go
func (nsexec *Executor) Execute(binary string, args []string, timeout time.Duration) (string, error) {
	return nsexec.executor.Execute(nil, types.NsBinary, nsexec.prepareCommandArgs(binary, args), timeout)
}
```

To:
```go
func (nsexec *Executor) Execute(envs []string, binary string, args []string, timeout time.Duration) (string, error) {
	return nsexec.executor.Execute(nil, types.NsBinary, nsexec.prepareCommandArgs(binary, args, envs), timeout)
}
```

So, we need to bump following dependencies:

- Bump Harvester to v1.3.1
- Replace go-common-libs

## Test Plan

1. Generate binary

```shell
CGO_ENABLED=0 go build -mod=vendor -o bin/terraform-provider-harvester .
mkdir -p ~/.terraform.d/plugins/registry.terraform.io/harvester/harvester/0.0.0-dev/darwin_arm64
cp bin/terraform-provider-harvester ~/.terraform.d/plugins/registry.terraform.io/harvester/harvester/0.0.0-dev/darwin_arm64/terraform-provider-harvester_v0.0.0-dev
```

2. Prepare a provider.tf

```tf
terraform {
  required_providers {
    harvester = {
      source = "harvester/harvester"
      version = "0.0.0-dev" 
    }
  }
}

provider "harvester" {
    kubeconfig = "your harvester kube config"
}
```

3. Use following main.tf to test 

```tf
resource "harvester_image" "ubuntu20" {
   name      = "ubuntu-focal"
   namespace = "default"

   display_name = "ubuntu-focal"
   source_type  = "download"
   url          = "https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-amd64.img"
 }

resource "harvester_cloudinit_secret" "cloud-config-ubuntu20" {
  name      = "cloud-config-ubuntu20"
  namespace = "default"

  user_data    = <<-EOF
    #cloud-config
    password: test
    chpasswd:
      expire: false
    ssh_pwauth: true
    package_update: true
    packages:
      - qemu-guest-agent
    runcmd:
      - - systemctl
        - enable
        - '--now'
        - qemu-guest-agent
    EOF
  network_data = ""
}

resource "harvester_virtualmachine" "ubuntu20" {
  name                 = "ubuntu20"
  namespace            = "default"
  restart_after_update = true

  description = "test ubuntu20 raw image"
  tags = {
    ssh-user = "ubuntu"
  }

  cpu    = 1
  memory = "2Gi"

  hostname        = "ubuntu20"
  reserved_memory = "100Mi"
  machine_type    = "q35"

  network_interface {
    name           = "nic-1"
    wait_for_lease = true
  }

  disk {
    name       = "rootdisk"
    type       = "disk"
    size       = "10Gi"
    bus        = "virtio"
    boot_order = 1

    image       = harvester_image.ubuntu20.id
    auto_delete = true
  }

  cloudinit {
    user_data_secret_name    = harvester_cloudinit_secret.cloud-config-ubuntu20.name
    network_data_secret_name = harvester_cloudinit_secret.cloud-config-ubuntu20.name
  }
}
```


bk201 (approved):
LGTM, thanks!

FrankYang0529 (approved):
LGTM. Thank you. Test with goreleaser:

1. Add a temporary tag.
```
git tag v0.6.6
```
2. Remove `dist` folder.
```
rm -rf dist
```
3. Build.
```
> goreleaser build
  • loading environment variables
  • getting and validating git state
    • git state                                      commit=ec81b3b84818d61fd663d8ee52bf934e7be5934a branch=PR-106 current_tag=v0.6.6 previous_tag=v0.6.5 dirty=false
  • parsing tag
  • setting defaults
  • running before hooks
    • running                                        hook=go mod tidy
  • ensuring distribution directory
  • setting up metadata
  • writing release metadata
  • loading go mod information
  • build prerequisites
  • building binaries
    • building                                       binary=dist/terraform-provider-harvester_darwin_arm64/terraform-provider-harvester_v0.6.6
    • building                                       binary=dist/terraform-provider-harvester_darwin_amd64_v1/terraform-provider-harvester_v0.6.6
    • building                                       binary=dist/terraform-provider-harvester_linux_amd64_v1/terraform-provider-harvester_v0.6.6
    • building                                       binary=dist/terraform-provider-harvester_linux_arm64/terraform-provider-harvester_v0.6.6
    • took: 54s
  • writing artifacts metadata
  • build succeeded after 54s
  • thanks for using goreleaser!
```

FrankYang0529:
Hi @m-ildefons, could you help to review this? Thank you.

m-ildefons (approved):
LGTM.

I have no Darwin system to test this change on, but it looks sane to me.

I've looked at why we even depend on importing longhorn in the first place and it looks to me like the only place where we utilize the Longhorn library is in the storage class resource schema. On top of that, the only thing from the Longhorn library we use is the constant `LonghornDriverName` as the default value for the provider for the storage class resource. IMO, we should hard-code this as a local constant and remove the dependency on the Longhorn library completely. What do you think?

Yu-Jack:
> LGTM.
> 
> I have no Darwin system to test this change on, but it looks sane to me.
> 
> I've looked at why we even depend on importing longhorn in the first place and it looks to me like the only place where we utilize the Longhorn library is in the storage class resource schema. On top of that, the only thing from the Longhorn library we use is the constant `LonghornDriverName` as the default value for the provider for the storage class resource. IMO, we should hard-code this as a local constant and remove the dependency on the Longhorn library completely. What do you think?

I agree with it. If we all agree with this one, I could remove it in this PR until we really need it. What do you think? @bk201 @FrankYang0529

bk201:
Sound good it the constant is never expected to change.

Yu-Jack:
@m-ildefons I just realized because Harvester v1.3.1 still uses LH as dependency, so removing LH dependency doesn't have too much effect. So, I'll just merge this PR. cc @bk201 @FrankYang0529

go.mod
```diff
@@ -1,8 +1,11 @@
 module github.com/harvester/terraform-provider-harvester
 
-go 1.22
+go 1.22.5
 
 replace (
+	// From this commiet, it starts to support Darwin.
+	github.com/longhorn/go-common-libs => github.com/longhorn/go-common-libs v0.0.0-20240926084818-3a320d860af4
```
innobead, line 7: Why do we need to specially have this replace directive? Don't we use the latest commit of go-common-libs as it's an indirect dependency and has the darwin arm64 support already?

@Yu-Jack

Yu-Jack: @innobead  Actually, at that moment, we were using latest commit. This PR was open on Sep 27, 2024. The next commit was made on Oct 5, 2024. We had not made any commits when we merged this PR.

image

The reason we need this replacement is because of dependency chain. You can't directly remove the replacement, it results some dependency errors. In the meantime, you can't use latest commit either. Because latest go-common libs might be only designed to latest longhorn instead of old one.

Hence, if we'd like to bump this one, we need to spend time on checking dependency chain.

PR #108, opened 2024-10-22
Bump Go Dependencies

m-ildefons:
- Migrate to Wrangler v3
- Remove explicit dependency on Longhorn-manager
- Bump Harvester to v1.3.2

As discussed in https://github.com/harvester/terraform-provider-harvester/pull/106, I've removed the explicit dependency on Longhorn-manager by replacing the imported constant string with a local constant string.


Yu-Jack (approved):
LGTM

FrankYang0529 (approved):

PR #113, opened 2025-01-03
Add Volume Delete Test Step

m-ildefons:
Add test step ensuring that volumes are deleted correctly. This test works independent of the timeout since in case the timeout is exceeded the delete operation will already be considered a fail. Thus, if the test step succeeds, the delete would not have timed out. The test step also ensures that the resource is actually removed.

The actual delete logic for volumes was already fixed in 55ee9e6 (https://github.com/harvester/terraform-provider-harvester/pull/103)

related-to: harvester/harvester#6415


FrankYang0529 (approved):
LGTM. Thanks for the patch.

brandboat (commented):
LGTM, thanks!

brandboat (approved):

bk201 (approved):

PR #114, opened 2025-01-06
Terraform Provider: Kubeconfig as Base64 string

m-ildefons:
Accept kubeconfig as base64 string.
When configuring the Harvester Terraform provider, the kubeconfig can either be given as path to a file, from which to read, or the contents can be supplied directly as base64 string.
This allows for greater flexibility in supplying this sensitive data to the Terraform provider.

related-to: harvester/harvester#6234

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -3,6 +3,8 @@ package client
 import (
 	"encoding/base64"
 
+	"github.com/mitchellh/go-homedir"
+
 	harvnetworkclient "github.com/harvester/harvester-network-controller/pkg/generated/clientset/versioned"
 	harvclient "github.com/harvester/harvester/pkg/generated/clientset/versioned"
 	"github.com/harvester/harvester/pkg/generated/clientset/versioned/scheme"
@@ -70,7 +72,12 @@ func NewClient(kubeConfig, kubeContext string) (*Client, error) {
 }
 
 func restConfigFromFile(kubeConfig, kubeContext string) (*rest.Config, error) {
-	clientConfig := kubeconfig.GetNonInteractiveClientConfigWithContext(kubeConfig, kubeContext)
+	clientConfigPath, err := homedir.Expand(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	clientConfig := kubeconfig.GetNonInteractiveClientConfigWithContext(clientConfigPath, kubeContext)
 	return clientConfig.ClientConfig()
 }
 
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1,6 +1,8 @@
 package client
 
 import (
+	"encoding/base64"
+
 	harvnetworkclient "github.com/harvester/harvester-network-controller/pkg/generated/clientset/versioned"
 	harvclient "github.com/harvester/harvester/pkg/generated/clientset/versioned"
 	"github.com/harvester/harvester/pkg/generated/clientset/versioned/scheme"
@@ -9,6 +11,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	storageclient "k8s.io/client-go/kubernetes/typed/storage/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 type Client struct {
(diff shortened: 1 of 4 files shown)


ihcsim:
Are we ok with the cluster user private key (`users.user[*].client-key-data`) being stored as plain text in the Terraform's state file? I am tempted to say we shouldn't support anything more than what the [Terraform Kubernetes provider supports](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#authentication). Anything more feels like user CI-specific problem.

m-ildefons:
This is indeed a concern, but Terraform Enterprise or HCP Terraform are able to encrypt sensitive data before storing it:
https://developer.hashicorp.com/terraform/tutorials/configuration-language/sensitive-variables#sensitive-values-in-state

ihcsim:
If it's up to the user to secure their statefile while using our provider, then I think we should document it somewhere. Normally, a provider would generate sensitive data as an `output` or `data`. Then the user can decide if they want to write them to disk or use in memory.

JasonIamAUnixAdmin:
This would help us with an issue we have getting this to work with Terraform cloud

sdt-ssexton:
Looking forward to this as I think this would fix an issue we're encountering

m-ildefons:
I've added a warning in the documentation/example that informs the reader of the risk of leaking sensitive information via the Terraform state files.

FrankYang0529 (approved):
LGTM. I can use `kubeconfig` with base64 encode. Thanks.

pkg/constants/constants.go
```diff
@@ -4,8 +4,9 @@ const (
 	NamespaceDefault         = "default"
 	NamespaceHarvesterSystem = "harvester-system"
 
-	FieldProviderKubeConfig  = "kubeconfig"
-	FieldProviderKubeContext = "kubecontext"
+	FieldProviderKubeConfig       = "kubeconfig"
+	FieldProviderKubeContext      = "kubecontext"
+	FieldProviderKubeConfigBase64 = "kubeconfig_base64"
```
ihcsim, line 9: It doesn't look like this is being used.

pkg/client/client.go
```diff
@@ -59,3 +68,16 @@ func NewClient(kubeConfig, kubeContext string) (*Client, error) {
 		HarvesterNetworkClient:    harvNetworkClient,
 	}, nil
 }
+
+func restConfigFromFile(kubeConfig, kubeContext string) (*rest.Config, error) {
```
ihcsim, line 72: Will it make sense to move the `homedir.Expand(d.Get(constants.FieldProviderKubeConfig).(string))` at https://github.com/harvester/terraform-provider-harvester/blob/35e2902d29e6590b80bb8e05b9c1af41f5aa1ca9/internal/provider/provider.go#L71-L74 into this new function?

pkg/constants/constants.go
ibrokethecloud: +1, we should remove this.

pkg/client/client.go
m-ildefons: Yeah. That would make sense.

`homedir.Expand` really on expands the `~` to whatever the invoking user's home directory path is. It's only relevant if you're dealing with a path, not if you're dealing with base64 encoded strings, since `~` can't occur in base64 data.

I wonder if there isn't a better more canonical way to do this only with stuff from the Golang standard library. We're taking on a dependency only for this and the dependency isn't maintained anymore. Not a big deal, since it's not code that's likely to blow up in our face anytime soon, but still not good.
But we do get some portability for MacOS and Windows from this library, so ripping it out isn't necessarily trivial - depending on what the state of portability for Golang is nowadays.

pkg/constants/constants.go
m-ildefons: Removed. Good catch.

ihcsim (commented):
LGTM - thanks for working on this.

ihcsim (approved):

pkg/client/client.go
ihcsim: Yeah, I am not fond of that dependency. I'm betting it came from some terraform example code. Personally, I will just make it an absolute path, and enforce it with `filepath.IsAbs()`.

ibrokethecloud (approved):

PR #118, opened 2025-03-25
feat: add setting to terraform

FrankYang0529:
https://github.com/harvester/harvester/issues/4354

## Tests

1. Setup backup target and wait few miniutes. The NFS needs time to start.

```
kubectl apply -f https://raw.githubusercontent.com/longhorn/longhorn/refs/tags/v1.7.2/deploy/backupstores/nfs-backupstore.yaml
```

2. Use terraform to update setting.

```tf
resource "harvester_setting" "backup-target" {
  name  = "backup-target"
  value = jsonencode(
    {
      endpoint = "nfs://longhorn-test-nfs-svc.default:/opt/backupstore"
      type     = "nfs"
    }
  )
}
```

3. Check setting has been updated.
```
kubectl get setting backup-target -o yaml
```

4. Change value to empty string and use terraform to update setting. Check the setting is reset.

```tf
resource "harvester_setting" "backup-target" {
  name  = "backup-target"
  value = ""
}
```

5. Remove backup-target in the terraform and run `terraform apply` again. Check the setting is not removed. It just resets the value to empty.


m-ildefons (commented):
What about other settings? This seems to implement the settings resource in general, but the test and example seems to only be concerned with the backup-target setting.
I see no reason why it wouldn't work for other settings, but if this is a general settings resource, then it should be made clear that all settings are in scope and must work.

Since settings can't be created and deleted like normal resources, I would really like to see tests here to make sure that behavior is always correctly kept.

FrankYang0529:
> I see no reason why it wouldn't work for other settings, but if this is a general settings resource, then it should be made clear that all settings are in scope and must work.

Why do you think it cannot work? I just use backup-target as an example. Feel free to use other setting to test it. I already done some validation.

> Since settings can't be created and deleted like normal resources, I would really like to see tests here to make sure that behavior is always correctly kept.

It's already covered by test steps, right? When you first time run `terraform apply`, it calls creation because there is no data in the state. When you remove data from the terraform, it calls deletion.

m-ildefons:
> It's already covered by test steps, right? When you first time run terraform apply, it calls creation because there is no data in the state. When you remove data from the terraform, it calls deletion.

I meant adding automated tests in `internal/tests/`.

FrankYang0529:
> I meant adding automated tests in `internal/tests/`.

Added it.

m-ildefons (approved):
LGTM

internal/provider/setting/resource_setting.go
```diff
@@ -0,0 +1,141 @@
...
+	toUpdate, err := util.ResourceConstruct(d, Updater(obj))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	_, err = c.HarvesterClient.HarvesterhciV1beta1().Settings().Update(ctx, toUpdate.(*harvsterv1.Setting), metav1.UpdateOptions{})
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return resourceSettingRead(ctx, d, meta)
```
brandboat, line 87: Does the read operation here is necessary? I'm wondering could we invoke `diag.FromErr(resourceSettingImport(d, obj))` here?

internal/provider/setting/resource_setting.go
```diff
@@ -0,0 +1,141 @@
...
+		if apierrors.IsNotFound(err) {
+			d.SetId("")
+			return nil
+		}
+		return diag.FromErr(err)
+	}
+	return diag.FromErr(resourceSettingImport(d, obj))
+}
+
+// The setting cannot be deleted. It can only be resetted to empty.
```
brandboat, line 107: How about we improve the comment a bit?
`// The setting cannot be deleted; it can only be reset to an empty string (""), which represents using the default value.`

brandboat (dismissed):
LGTM % minor comments. 🚀🚀🚀  

Verify steps
---
- Run ACC tests and corresponding tests passed.
  image

- Manual Test: follow the test steps in PR desc, and the backup-target setting updated as expected.
`terraform apply`
  ```
  resource "harvester_setting" "backup-target" {
    name  = "backup-target"
    value = jsonencode(
      {
        endpoint = "nfs://longhorn-test-nfs-svc.default:/opt/backupstore"
        type     = "nfs"
      }
    )
  }
  ```
  image
- Test update/delete setting, all work as expected.

internal/provider/setting/resource_setting.go
FrankYang0529: Yes, we can just use updated obj for `diag.FromErr(resourceSettingImport(d, obj))`. Thanks.

brandboat (approved):

bk201 (approved):

PR #123, opened 2025-06-03
feat: add harvester support matrix

FrankYang0529:


Vicente-Cheng (approved):
thanks for the improvement!

PR #133, opened 2025-08-11
add checksum field to image resource schema

m-ildefons:
Add checksum field to image resource schema and constructor. The checksum allows users to make sure an image from a download source is downloaded correctly and the content are as expected.

Changing the checksum field forces the image CRD to be re-created, because only this forces the image to be downloaded again and the checksum to be re-calculated and compared with the updated value.

related-to: harvester/harvester#8265

#### Problem:

Checksum field was absent

#### Solution:

Add checksum field

#### Related Issue(s):

harvester/harvester#8265

#### Test plan:

It's now possible to set the checksum of an image with type "download" when creating it with the Terraform provider:

```
resource "harvester_image" "opensuse_156" {
  name      = "opensuse-15.6"

  display_name = "openSUSE-Leap-15.6"
  source_type  = "download"
  url          = "http://192.168.0.254/artifacts/openSUSE-Leap-15.6.x86_64-NoCloud.qcow2"
  checksum     = "6508cc3df86a92a7051aed50349763da277801efc8bd18aa27ff13c4325bb49117233039e2f19fc5ec8fb4587a7d246d2161d37ba015946d1bbdb8b4500c64c2"

  storage_class_name = "harvester-longhorn"
}
```

Screenshot at 2025-08-11 15-24-42

### DIFF
--- a/docs/data-sources/image.md
+++ b/docs/data-sources/image.md
@@ -37,6 +37,7 @@ data "harvester_image" "opensuse154" {
 ### Read-Only
 
 - **backend** (String) The backend type of the image, either 'backing-image' or 'cdi'.
+- **checksum** (String) SHA-512 checksum of the image
 - **description** (String) Any text you want that better describes this resource
 - **message** (String)
 - **progress** (Number)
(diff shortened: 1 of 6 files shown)


FrankYang0529 (commented):
Please also update `resource_image_importer.go` file.

FrankYang0529 (approved):

PR #139, opened 2025-09-04
ci: use release event to avoid ci skip [ci skip]

starbops:
**Problem:**

If the tagged commit contains any CI-skip requests, the release workflow cannot be triggered.

**Solution:**

Use the `release` event instead of the `push` event. Ideally, both prereleases and releases can trigger the workflow no matter what forms of CI-skip requests are made in the commit message.

**Related Issue:**

harvester/harvester#9048

**Test plan:**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,9 +11,9 @@
 #
 name: release
 on:
-  push:
-    tags:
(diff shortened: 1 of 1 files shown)


starbops:
@mergify backport v1.6

PR #190, opened 2026-03-18
Feat: PCI Device Support

m-ildefons:
## Problem:

The Harvester Terraform provider is missing several key features, which are necessary to be able to manage virtual machines with host PCI devices passed through to the guest.

- The `harvester_virtualmachine` resource is missing the necessary options to configure host PCI devices.
- There is no feature in the Harvester Terraform provider, which can be used to enable a host PCI device for PCI passthrough
- There is no feature in the Harvester Terraform provider to manage SR-IOV devices and configure the number of virtual functions (VFs)

## Solution:

- Add SR-IOV network device data source and resource

The SR-IOV network device resource is used to configure the number for virtual functions (VFs) for that partilcular device.

The data source can be used to query the device for the PCI device names of the virtual functions.

- Add a data source and resource for PCI Devices.

This data source can be used to query the Harvester cluster for information about specific PCI devices, such as the address, vendor and device IDs.

The resource can be used to enable/disable PCI passthrough for that particular PCI device.

- Add a host_device block to the harvester_virtualmachine resource.

This block can be used to attach a PCI Device, which has been enabled for PCI passthrough, to a VM.

## Related Issue(s):

related-to: https://github.com/harvester/harvester/issues/6423

### Prerequisites

- A Harvester cluster (v1.7.1 or newer) with SR-IOV capable network devices
- Harvester PCI device addon enabled

### Terraform provider build and installation

1) Clone/checkout the PR branch
```
git clone git@github.com:m-ildefons/terraform-provider-harvester.git
cd terraform-provider-harvester
git checkout gh-6423
```

2) Build the provider binary

In the Terraform provider git repo:
```
make build
```

3) Install the provider

In the Terraform provider git repo:
```
mkdir -p ~/.terraform.d/plugins/terraform.local/local/harvester/0.0.0-dev/linux_amd64/
cp bin/terraform-provider-harvester-amd64 ~/.terraform.d/plugins/terraform.local/local/harvester/0.0.0-dev/linux_amd64/terraform-provider-harvester_v0.0.0-dev
terraform init
```

### Set up for testing

- Create a test directory
- Download the Kubeconfig file of the Harvester cluster into this test directory and name it `kubeconfig_test.yaml`
- Set up a Terraform file for using the provider in the test directory:
```
terraform {
  required_providers {
    harvester = {
      source  = "terraform.local/local/harvester"
      version = "0.0.0-dev"
    }
  }
}

provider "harvester" {
  kubeconfig = "./kubeconfig_test.yaml"
}
```
- Append Terraform resources for the SR-IOV network device and the PCI devices. See examples below.
- To create/update resources, run `terraform apply`
- To remove resources, run `terraform destroy`
- To create/update/remove only specific resources use `terraform apply -target resource_type.resource_name`

#### Example 1 - Multiple VFs, multiple VMs with one VF each

```terraform
# Configure root PCI device, enable several VFs.
# The names of the resulting PCI devices of the VFs will be in the property
# "vf_device_names" of this resource, after it has been configured
resource "harvester_sriov_network_device" "hp-49-eno50-root" {
  name = "hp-49-tink-system-eno50"  # name of the root PCI device to configure
  virtual_functions = 2
}

# Enable PCI passthrough for each VF
resource "harvester_pcidevice" "hp-49-eno50-vfs" {
  depends_on = [
    harvester_sriov_network_device.hp-49-eno50-root
  ]
  
  for_each = harvester_sriov_network_device.hp-49-eno50-root.vf_device_names
  
  name = each.key
  passthrough_enabled = true
}

resource "harvester_image" "ubuntu24" {
  name         = "ubuntu24"
  namespace    = "harvester-public"
  display_name = "ubuntu-24.04"
  source_type  = "download"
  url          = "https://cloud-images.ubuntu.com/releases/releases/24.04/release/ubuntu-24.04-server-cloudimg-amd64.img"
}

# For each VF create a separate VM. Pass only that one VF through to that VM.
resource "harvester_virtualmachine" "vm" {
  depends_on = [
    harvester_sriov_network_device.hp-49-eno50-root,
    harvester_pcidevice.hp-49-eno50-vfs,
  ]
  
  for_each = harvester_sriov_network_device.hp-49-eno50-root.vf_device_names
  
  name = "vm-${each.key}"
  namespace = "default"
  description = "Test VM for PCI device ${each.key}"
  
  cpu    = 2
  memory = "4Gi"
  
  reserved_memory = "100Mi"
  machine_type    = "q35"
  hostname        = "vm-${each.key}"
  
  cloudinit {
    user_data = <<-EOF
      #cloud-config
      user: ubuntu
      password: foobar
      chpasswd:
        expire: false
      package_update: true
      packages:
        - qemu-guest-agent
      runcmd:
        - - systemctl
          - enable
          - --now
          - qemu-guest-agent.service
    EOF
  }
  
  disk {
    name       = "rootdisk"
    type       = "disk"
    size       = "10Gi"
    bus        = "virtio"
    boot_order = 1
    image       = harvester_image.ubuntu24.id
    auto_delete = true
  }
  
  host_device {
    name        = each.key
    device_name = harvester_pcidevice.hp-49-eno50-vfs[each.key].resource_name
  } 
}
```

#### Example 2 - Multiple VFs, one VM with all VFs

```terraform
# Configure several VFs on an SR-IOV PCI device
resource "harvester_sriov_network_device" "hp-49-eno50-root" {
  name = "hp-49-tink-system-eno50"  # name of the root PCI device to configure
  virtual_functions = 2
}

# Enable passthrough for each VF
resource "harvester_pcidevice" "hp-49-eno50-vfs" {
  depends_on = [
    harvester_sriov_network_device.hp-49-eno50-root
  ]
  
  for_each = harvester_sriov_network_device.hp-49-eno50-root.vf_device_names
  
  name = each.key
  passthrough_enabled = true
}

resource "harvester_image" "ubuntu24" {
  name         = "ubuntu24"
  namespace    = "harvester-public"
  display_name = "ubuntu-24.04"
  source_type  = "download"
  url          = "https://cloud-images.ubuntu.com/releases/releases/24.04/release/ubuntu-24.04-server-cloudimg-amd64.img"
}

# Create one VM, pass all VFs through to that VM.
resource "harvester_virtualmachine" "vm" {
  depends_on = [
    harvester_sriov_network_device.hp-49-eno50-root,
    harvester_pcidevice.hp-49-eno50-vfs,
  ]
  
  name = "vm-multiple-vfs"
  namespace = "default"
  description = "Test VM for multiple PCI devices"
  
  cpu    = 2
  memory = "4Gi"
  
  reserved_memory = "100Mi"
  machine_type    = "q35"
  hostname        = "multidevice"
  
  cloudinit {
    user_data = <<-EOF
      #cloud-config
      user: ubuntu
      password: foobar
      chpasswd:
        expire: false
      package_update: true
      packages:
        - qemu-guest-agent
      runcmd:
        - - systemctl
          - enable
          - --now
          - qemu-guest-agent.service
    EOF
  }
  
  disk {
    name       = "rootdisk"
    type       = "disk"
    size       = "10Gi"
    bus        = "virtio"
    boot_order = 1
    image       = harvester_image.ubuntu24.id
    auto_delete = true
  }
  
  # Use Terraform dynamic blocks to create multiple "host_device" blocks in
  # "harvester_virtualmachine" resource.
  dynamic "host_device" {
    for_each = harvester_sriov_network_device.hp-49-eno50-root.vf_device_names

    content {
      name = host_device.key
      device_name = harvester_pcidevice.hp-49-eno50-vfs[host_device.key].resource_name
    }
  }
}
```


internal/provider/pcidevice/datasource_pcidevice.go
```diff
@@ -0,0 +1,43 @@
...
+	return diag.FromErr(resourcePCIDeviceImport(d, pcidevice))
+}
+
+func resourcePCIDeviceImport(d *schema.ResourceData, obj *devicesv1.PCIDevice) error {
+	stateGetter, err := importer.ResourcePCIDeviceStateGetter(obj)
+	if err != nil {
+		return err
+	}
+	return util.ResourceStatesSet(d, stateGetter)
+}
```
brandboat: Let's put this method to `resource_pcidevice.go` for consistency with other resources, e.g. vm, volume...

internal/provider/pcidevice/resource_pcidevice.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	name := d.Get(constants.FieldCommonName).(string)
+	enabled := d.Get(constants.FieldPCIDevicePassthroughEnabled).(bool)
+	obj, err := c.HarvesterDeviceClient.DevicesV1beta1().PCIDevices().Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			d.SetId("")
+			return nil
```
brandboat: Do we need to return error if pcidevice does not exist?

internal/provider/pcidevice/resource_pcidevice.go
```diff
@@ -0,0 +1,214 @@
...
+								Kind:       "PCIDevice",
+								Name:       name,
+								UID:        obj.UID,
+							},
+						},
+					},
+					Spec: devicesv1.PCIDeviceClaimSpec{
+						Address:  obj.Status.Address,
+						NodeName: obj.Status.NodeName,
+						UserName: "admin",
```
brandboat, line 77: Perhaps we can retrieve username from kubeconfig instead of hardcoding it to "admin"?  I'm really not sure what's the best way to handle the username here... c.c. @Yu-Jack

internal/provider/pcidevice/resource_pcidevice.go
```diff
@@ -0,0 +1,214 @@
...
+					},
+				}
+				_, err = c.HarvesterDeviceClient.DevicesV1beta1().PCIDeviceClaims().Create(ctx, toCreate, metav1.CreateOptions{})
```
brandboat, line 80: Do we need to invoke `resourcePCIDeviceWaitForState` after creating PCIDeviceClaim? We should wait until StatePassthroughEnabled is enable just like L87 did, right?

brandboat (commented):
Thanks for the feature! I just had a quick look.

internal/provider/pcidevice/resource_pcidevice.go
m-ildefons: I think you're right. This logic is switched up.
Waiting for the creation should only happen if we actually create a PCIDeviceClaim. Otherwise, if there was no error, we can just import the device state. On error, we need to return with that error of course.

internal/provider/pcidevice/resource_pcidevice.go
m-ildefons: I put a hardcoded `admin` here, because I believe the username field `.spec.username` is not actually used anywhere in the PCI device controller.
If this field does have a function, then I agree, it would be best to just extract the username from the kubeconfig.

internal/provider/pcidevice/resource_pcidevice.go
m-ildefons: Yes, good spot.

internal/provider/sriovdevice/datasource_sriovnetworkdevice.go
```diff
@@ -0,0 +1,43 @@
...
+	return diag.FromErr(resourceSRIOVNetworkDeviceImport(d, pcidevice))
+}
+
+func resourceSRIOVNetworkDeviceImport(d *schema.ResourceData, obj *devicesv1.SRIOVNetworkDevice) error {
+	stateGetter, err := importer.ResourceSRIOVNetworkDeviceStateGetter(obj)
+	if err != nil {
+		return err
+	}
+	return util.ResourceStatesSet(d, stateGetter)
+}
```
brandboat: Like https://github.com/harvester/terraform-provider-harvester/pull/190#discussion_r3026453059, let's put this under `resource_sriovnetworkdevice.go`

internal/provider/pcidevice/resource_pcidevice.go
```diff
@@ -0,0 +1,235 @@
...
+
+func resourcePCIDeviceWaitForState(ctx context.Context, d *schema.ResourceData, meta interface{}, timeoutKey string) error {
+	var (
+		pending []string
+		target  []string
+	)
+
+	if timeoutKey == schema.TimeoutDelete {
+		pending = []string{constants.StatePassthroughEnabled}
+		target = []string{constants.StatePassthroughDisabled}
```
brandboat, line 176: For the delete case, do we need to add `constants.StateCommonRemoved` to the target? Otherwise we need to wait until timeout if the pcidevice not exists.

internal/provider/pcidevice/resource_pcidevice.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+
+		claims, err := c.HarvesterDeviceClient.DevicesV1beta1().PCIDeviceClaims().List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return obj, constants.StateCommonError, err
+		}
+
+		for _, claim := range claims.Items {
+			for _, owner := range claim.OwnerReferences {
+				if owner.Name == name && claim.Status.PassthroughEnabled {
```
brandboat, line 219: Couldn't we retrieve the PCIDeviceClaim through Get method: `c.HarvesterDeviceClient.DevicesV1beta1().PCIDeviceClaims().Get(ctx, name, metav1.GetOptions{})`? From the `resourcePCIDeviceCreate`, I assume the PCIDeviceClaim name & ownerreference name is the same as PCIDevice name.

internal/provider/sriovdevice/resource_sriovnetworkdevice_constructor.go
```diff
@@ -0,0 +1,59 @@
...
+	return c.SRIOVNetworkDevice, nil
+}
+
+func newSRIOVNetworkDeviceConstructor(SRIOVNetworkDevice *devicesv1.SRIOVNetworkDevice) util.Constructor {
+	return &Constructor{
+		SRIOVNetworkDevice: SRIOVNetworkDevice,
+	}
+}
+
+func Creator(name string) util.Constructor {
```
brandboat, line 50: Unused method, do we missed anything here?

internal/provider/sriovdevice/resource_sriovnetworkdevice.go
```diff
@@ -0,0 +1,152 @@
...
+	c, err := meta.(*config.Config).K8sClient()
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	name := d.Get(constants.FieldCommonName).(string)
+	obj, err := c.HarvesterDeviceClient.DevicesV1beta1().SRIOVNetworkDevices().Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			d.SetId("")
+			return nil
```
brandboat: Do we need to return error of device not found?

internal/provider/sriovdevice/resource_sriovnetworkdevice.go
```diff
@@ -0,0 +1,152 @@
...
+
+func resourceSRIOVNetworkDeviceWaitForState(ctx context.Context, d *schema.ResourceData, meta interface{}, timeoutKey string) error {
+	var (
+		pending []string
+		target  []string
+	)
+
+	if timeoutKey == schema.TimeoutDelete {
+		pending = []string{devicesv1.DeviceEnabled}
+		target = []string{devicesv1.DeviceDisabled}
```
brandboat, line 104: Do we need to add StateCommonRemoved to the target?

internal/provider/sriovdevice/resource_sriovnetworkdevice.go
```diff
@@ -0,0 +1,152 @@
...
+			if apierrors.IsNotFound(err) {
+				return obj, constants.StateCommonRemoved, nil
+			}
+			return obj, constants.StateCommonError, err
+		}
+		if err = resourceSRIOVNetworkDeviceImport(d, obj); err != nil {
+			return obj, constants.StateCommonError, err
+		}
+
+		state_raw := d.Get(constants.FieldSRIOVNetworkDeviceEnabled).(bool)
```
brandboat: nit: `stateRaw`

internal/provider/sriovdevice/resource_sriovnetworkdevice_constructor.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/harvester/terraform-provider-harvester/pkg/constants"
+)
+
+var (
+	_ util.Constructor = &Constructor{}
+)
+
+type Constructor struct {
+	SRIOVNetworkDevice *devicesv1.SRIOVNetworkDevice
+}
```
brandboat, line 16: Do we need to implement Constructor for pcidevice?

internal/provider/virtualmachine/schema_virtualmachine.go
```diff
@@ -181,6 +181,25 @@ please use %s instead of this deprecated field:
...
+		constants.FieldVirtualMachineHostDevice: {
+			Type:        schema.TypeList,
+			Description: "",
+			Optional:    true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					constants.FieldHostDeviceName: {
+						Type:        schema.TypeString,
+						Optional:    true,
+						Description: "",
```
brandboat: nit: Could you please add the desc here, or we can remove it.

internal/provider/virtualmachine/schema_virtualmachine.go
```diff
@@ -181,6 +181,25 @@ please use %s instead of this deprecated field:
...
+					},
+					constants.FieldHostDeviceDeviceName: {
+						Type:        schema.TypeString,
+						Optional:    true,
+						Description: "",
```
brandboat: ditto.

internal/provider/virtualmachine/schema_virtualmachine.go
```diff
@@ -181,6 +181,25 @@ please use %s instead of this deprecated field:
 			Default:     false,
 			Description: "Create an initial snapshot named {vm-name}-initial after the VM is created and ready",
 		},
+		constants.FieldVirtualMachineHostDevice: {
+			Type:        schema.TypeList,
+			Description: "",
```
brandboat: nit: Could you please add the desc here, or we can remove it.

internal/provider/pcidevice/resource_pcidevice.go
m-ildefons: The PCIDevice CR does not get removed when PCI passthrough is disabled for that particular device. Afterall, the device itself also doesn't disappear.

In my testing, the most reliable indicator for PCI passthrough is whether or not a PCIDeviceClaim CR for that PCIDevice CR exists.

internal/provider/pcidevice/resource_pcidevice.go
m-ildefons: AFAIK, there is nothing that guarantees that the PCIDeviceClaim has the same name as the PCIDevice. They are linked through ownership reference only.

internal/provider/sriovdevice/resource_sriovnetworkdevice.go
m-ildefons: Again, the lifetime of the SRIOVNetworkDevice CR is tied to the physical device and is managed by the PCIDevice controller. Therefore, the Terraform provider only manages configuration, but not creation or deletion of the CR.

internal/provider/sriovdevice/resource_sriovnetworkdevice_constructor.go
m-ildefons: Yes, the `Constructor` interface contains the conversion between the Terraform schema and the `devices.harvesterhci.io/v1beta1` Golang types.

internal/provider/pcidevice/resource_pcidevice.go
brandboat: Fair point, thanks for the explanation.

internal/provider/pcidevice/resource_pcidevice.go
brandboat: Got it, that means we need to change the logic in resourcePCIDeviceCreate, right?

We get PCIDeviceClaim by PCIDevice name in resourcePCIDeviceCreate

```go
	_, err = c.HarvesterDeviceClient.DevicesV1beta1().PCIDeviceClaims().Get(ctx, name, metav1.GetOptions{})
```

internal/provider/sriovdevice/resource_sriovnetworkdevice_constructor.go
brandboat: I only saw Constructor for SRIOVNetworkDevice, but seems we didn't implement it for PCIDevice, could you please help implement that?

brandboat (commented):
The PR is in good shape, but I think we still need some tweaks. @m-ildefons could you please address below 3 comments?
https://github.com/harvester/terraform-provider-harvester/pull/190#discussion_r3056621230 
https://github.com/harvester/terraform-provider-harvester/pull/190#discussion_r3050413029 
https://github.com/harvester/terraform-provider-harvester/pull/190#discussion_r3061558496

c.c @Yu-Jack, do you want to take another look?

Yu-Jack:
> c.c @Yu-Jack, do you want to take another look?

Sure, I'll take a look. Thanks for reminding.

internal/provider/sriovdevice/resource_sriovnetworkdevice_constructor.go
m-ildefons: I've re-implemented the resource conversion using the `Constructor` interface.
It wasn't as straight forward as I thought, but it revealed a few misconceptions of mine about how the `Constructor` interface works and consequently a few edge cases now work quite nicely.

Yu-Jack (commented):
Thanks for the PR. 

I have a question, what made you choose this way to enable pcidevice?

```
resource "harvester_pcidevice" "test" {
  name = "harvester1-000000010"
  passthrough_enabled = false
}
```

Instead of:

```
resource "harvester_pcidevice_claim" "test" {
  pcidevice_name = "harvester1-000000010"
}
```

I was considering whether one-one mapping is a suitable way for this like using `harvester_pcidevice_claim` to represent a enabled pci device as what we're doing in pcidevices-controller, which might be Intuitive to our users.

---

I also found a bug with steps:

1. terraform apply 
2. terraform destroy
3. complain nil pointer exception

```tf
data "harvester_pcidevice" "test" {
  name = "harvester1-000000010"
}

resource "harvester_pcidevice" "test" {
  name = "harvester1-000000010"
  passthrough_enabled = false
}
```

```
Stack trace from the terraform-provider-harvester_v0.0.0-dev plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x20 pc=0x1064626f0]

goroutine 87 [running]:
github.com/harvester/terraform-provider-harvester/internal/provider/pcidevice.resourcePCIDeviceDelete({0x1071ba390, 0x1400019f180}, 0x140001675f0, {0x106c5bd60, 0x140008099b0})
        /Users/xxxxx/terraform-provider-harvester/internal/provider/pcidevice/resource_pcidevice.go:127 +0x350
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).delete(0x14000728e00, {0x1071ba2e8, 0x14000787aa0}, 0x140001675f0, {0x106c5bd60, 0x140008099b0})
        /Users/xxxxx/terraform-provider-harvester/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema/resource.go:900 +0xe0
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).Apply(0x14000728e00, {0x1071ba2e8, 0x14000787aa0}, 0x1400083b5f0, 0x14000167560, {0x106c5bd60, 0x140008099b0})
        /Users/xxxxx/terraform-provider-harvester/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema/resource.go:949 +0x470
```

m-ildefons:
> I have a question, what made you choose this way to enable pcidevice?
> 
> ```
> resource "harvester_pcidevice" "test" {
>   name = "harvester1-000000010"
>   passthrough_enabled = false
> }
> ```
>
> Instead of:
>
> resource "harvester_pcidevice_claim" "test" {
>   pcidevice_name = "harvester1-000000010"
> }
>
> I was considering whether one-one mapping is a suitable way for this like using harvester_pcidevice_claim to represent a enabled pci device as what we're doing in pcidevices-controller, which might be Intuitive to our users.

That's a good point.
One-to-one mapping CRDs to Terraform resources is not always a great idea, especially if there is little or no documentation on the interactions of the different CRDs.
I'm trying to keep the workflow with the Terraform provider similar to the web UI. The web UI has at least a little bit of documentation and users will probably be fairly familiar with it anyways.
I don't think the PCIDeviceClaim CRD provides an "intuitive" UX at all.
Users will not be familiar with the internals of the pcidevices-controller and there is also no documentation at all for its CRDs. In fact without digging into the code and/or tracing the HTTP requests of the UI, it's not obvious from the existing workflows that a PCIDeviceClaim CRD even exists.

The Terraform provider needs at least a `harvester_pcidevice` data source, if not a resource anyways, because otherwise there is no way to query Harvester for properties of PCIDevice CRDs. This is necessary, because the PCIDevice CRD will contain the "resource name" property, which is later needed to assign PCI devices to VMs.
It would be a much more complicated workflow for users to have to deal with both PCIDevice and PCIDeviceClaim CRDs mirrored as Terraform resources, instead of having just one Terraform resource and let the Terraform provider handle managing the CRDs in the background.

Besides that, I'm really not quite satisfied with the current way PCIDeviceClaim CRDs are associated with the corresponding PCIDevice CRDs. There is no guarantee (to my knowledge) that the names of the CRDs will match between the PCIDevice and the PCIDeviceClaim.
Additionally, the names are also hard to predict for users, since they contain some information that isn't really accessible to them without looking at the system.
I would much prefer to do this association by PCI vendor/device ID or bus address, since these attributes are much more stable than the current approach with a name, which is a concatenation of different things.

Oh and thanks for pointing out that bug, I'll fix it.

Yu-Jack:
> That's a good point. One-to-one mapping CRDs to Terraform resources is not always a great idea, especially if there is little or no documentation on the interactions of the different CRDs. I'm trying to keep the workflow with the Terraform provider similar to the web UI. The web UI has at least a little bit of documentation and users will probably be fairly familiar with it anyways. I don't think the PCIDeviceClaim CRD provides an "intuitive" UX at all. Users will not be familiar with the internals of the pcidevices-controller and there is also no documentation at all for its CRDs. In fact without digging into the code and/or tracing the HTTP requests of the UI, it's not obvious from the existing workflows that a PCIDeviceClaim CRD even exists.
> 
> The Terraform provider needs at least a `harvester_pcidevice` data source, if not a resource anyways, because otherwise there is no way to query Harvester for properties of PCIDevice CRDs. This is necessary, because the PCIDevice CRD will contain the "resource name" property, which is later needed to assign PCI devices to VMs. It would be a much more complicated workflow for users to have to deal with both PCIDevice and PCIDeviceClaim CRDs mirrored as Terraform resources, instead of having just one Terraform resource and let the Terraform provider handle managing the CRDs in the background.
> 
> Besides that, I'm really not quite satisfied with the current way PCIDeviceClaim CRDs are associated with the corresponding PCIDevice CRDs. There is no guarantee (to my knowledge) that the names of the CRDs will match between the PCIDevice and the PCIDeviceClaim. Additionally, the names are also hard to predict for users, since they contain some information that isn't really accessible to them without looking at the system. I would much prefer to do this association by PCI vendor/device ID or bus address, since these attributes are much more stable than the current approach with a name, which is a concatenation of different things.

I got it. I'm good with the current approach based on your point. @ibrokethecloud do you have any concerns about this approach?

examples/resources/harvester_sriov_network_device/resource.tf
```diff
@@ -0,0 +1,4 @@
+resource "harvester_sriov_network_device" "eno50-root" {
```
ibrokethecloud: minor nit, should we include `hostname` in the resource name to ensure users get the right idea that similar to pcidevices these objects are tied to specific nodes

internal/provider/sriovdevice/resource_sriovnetworkdevice_constructor.go
```diff
@@ -0,0 +1,70 @@
...
+	processors := util.NewProcessors().
+		Tags(&c.SRIOVNetworkDevice.Labels).
+		Labels(&c.SRIOVNetworkDevice.Labels).
+		Description(&c.SRIOVNetworkDevice.Annotations)
+	customProcessors := []util.Processor{
+		{
+			Field: constants.FieldSRIOVNetworkDeviceNumVFs,
+			Parser: func(i interface{}) error {
+				num := i.(int)
+				tflog.Info(c.ctx, fmt.Sprintf("FOOBAR: %d", num))
```
ibrokethecloud: not sure if this is intentional

ibrokethecloud:
> Besides that, I'm really not quite satisfied with the current way PCIDeviceClaim CRDs are associated with the corresponding PCIDevice CRDs. There is no guarantee (to my knowledge) that the names of the CRDs will match between the PCIDevice and the PCIDeviceClaim. Additionally, the names are also hard to predict for users, since they contain some information that isn't really accessible to them without looking at the system. I would much prefer to do this association by PCI vendor/device ID or bus address, since these attributes are much more stable than the current approach with a name, which is a concatenation of different things.

technically `name` can be anything, users only need to be aware of `device_name` which needs to be matched to the resource name advertised by pcidevices + kubelet to the node

for example say a user cluster as 4 devices of same resourceType.. dev1,dev2,dev3,dev4

User enables all 4 dev's via terraform and passes `dev1/resourceType` to a VM.

Kubelet may actually assign `dev2` though terraform state will indicate `dev1`

In subsequent terraform apply the state will be used to generate the vm object which will reset name to `dev1` and VM will get an update.

During deletion scenario, user may try and disable/delete pcidevice `dev1` since they believe it is not in use, though in reality it may have been assigned to another VM outside of terraform control.

Harvester UI handles this better because we show the actual device in UI, which user can disable.

I see no real work around this at the moment until we make some fundamental changes to device plugin framework in pcidevices. For now we should just document this clearly so users can structure their pcidevice object calls separate from vm creation logic.

internal/provider/sriovdevice/resource_sriovnetworkdevice.go
```diff
@@ -0,0 +1,183 @@
...
+		} else {
+			state = devicesv1.DeviceDisabled
+		}
+		return obj, state, err
+	}
+}
+
+func countPCIDevices(ctx context.Context, client *client.Client, sriovNetworkDevice *devicesv1.SRIOVNetworkDevice) (int, error) {
+	numDevicesReady := 0
+	for _, pcideviceName := range sriovNetworkDevice.Status.VFPCIDevices {
```
ibrokethecloud: not really an issue, but pcidevices owned by sriovnetworkdevices contain a label `harvesterhci.io/parent-sriov-network-device: sriovNetworkDeviceName` to allow easy filtering the UI 

```
apiVersion: devices.harvesterhci.io/v1beta1
kind: PCIDevice
metadata:
  annotations:
    harvesterhci.io/pcideviceDriver: ixgbevf
  creationTimestamp: "2026-05-20T03:41:31Z"
  generation: 1
  labels:
    harvesterhci.io/parent-sriov-network-device: hp-130-tink-system-eno50
    nodename: hp-130-tink-system
  name: hp-130-tink-system-000004101
  resourceVersion: "7572639"
  uid: 0803f614-cb49-42dd-a40c-f7ad4a78d640
```

We could leverage the same to run a `list` and reconcile count with devices found in the list operation

ibrokethecloud (dismissed):

m-ildefons:
> > Besides that, I'm really not quite satisfied with the current way PCIDeviceClaim CRDs are associated with the corresponding PCIDevice CRDs. There is no guarantee (to my knowledge) that the names of the CRDs will match between the PCIDevice and the PCIDeviceClaim. Additionally, the names are also hard to predict for users, since they contain some information that isn't really accessible to them without looking at the system. I would much prefer to do this association by PCI vendor/device ID or bus address, since these attributes are much more stable than the current approach with a name, which is a concatenation of different things.
> 
> technically `name` can be anything, users only need to be aware of `device_name` which needs to be matched to the resource name advertised by pcidevices + kubelet to the node
> 
> for example say a user cluster as 4 devices of same resourceType.. dev1,dev2,dev3,dev4
> 
> User enables all 4 dev's via terraform and passes `dev1/resourceType` to a VM.
> 
> Kubelet may actually assign `dev2` though terraform state will indicate `dev1`
> 
> In subsequent terraform apply the state will be used to generate the vm object which will reset name to `dev1` and VM will get an update.
> 
> During deletion scenario, user may try and disable/delete pcidevice `dev1` since they believe it is not in use, though in reality it may have been assigned to another VM outside of terraform control.
> 
> Harvester UI handles this better because we show the actual device in UI, which user can disable.
> 
> I see no real work around this at the moment until we make some fundamental changes to device plugin framework in pcidevices. For now we should just document this clearly so users can structure their pcidevice object calls separate from vm creation logic.


That's true. `name` can be anything that is valid in a Kubernetes `ObjectMeta.Name` field.
But as far as I can tell, they are unique within a cluster, because PCIDevices are cluster scoped resources.
I'm not worried that `name` isn't well suited to identify a device.

However I am worried about predictability here, since names assigned by the PCI device controller seem to follow the schema `hostname-devicename`. However device names aren't really all that predictable. As you can see from the example, where the device name is `eno50` but that's probably just an alias for the real device name, which is something more akin to `enp0s3f6`.

The way around this would be to offer alternative methods for matching devices. E.g. based on hostname + PCI address or manufacturer and device IDs.
This would take away any possible ambiguity from device naming in Linux/Systemd and allow users to specify devices by properties that are certainly going to be unique.

My thought here is to make the name field optional and offer some additional optional fields like hostname, device ID, PCI bus address, manufacturer ID.
Then users can specify whichever and as many of these properties as they need to uniquely identify the device in the cluster.
The Terraform provider would then search the cluster for a device that matches all given properties. If there is none, or it's not unique, there can be an error.
But if there is only one device in the cluster matching all properties, then the Terraform provider can figure out the `.metadata.name` itself and continue to work. The user doesn't need to play around with `kubectl` or the web UI to identify each device they want to use with the Terraform provider.

internal/provider/sriovdevice/resource_sriovnetworkdevice_constructor.go
m-ildefons: Good spot. There is another one further down below as well ;-). I'll remove them.

examples/resources/harvester_sriov_network_device/resource.tf
m-ildefons: Fair enough. I just copied the example code from the .tf file I was testing with.
I'm happy to make it more obvious.

internal/provider/sriovdevice/resource_sriovnetworkdevice.go
m-ildefons: A `List()` is indeed better here, because it's only one API call no matter how many VFs a device has.
Even older ConnectX-3 NICs can support up to 126 VFs and since we're iterating on this every few seconds until the count matches the expected number of devices, this might spiral out of control.
With a `List()` we'd be making only a single API call every couple of seconds, which is reasonable when watching for resources to appear.

ibrokethecloud:
irrespective of what the `name` is it is not guaranteed to be honoured during VM pod creation device passthrough. I'd leave it as is for now as this is an issue that we need to address via the pcidevices controller

m-ildefons:
> irrespective of what the `name` is it is not guaranteed to be honoured during VM pod creation device passthrough. I'd leave it as is for now as this is an issue that we need to address via the pcidevices controller

You mean the `name` in the VM's `.Spec.Template.Spec.Domain.Devices.HostDevices`? I don't think that has any relvance at all. The examples from the Kubevirt docs just use `gpu1` and `gpu2`: https://kubevirt.io/user-guide/compute/host-devices/#starting-a-virtual-machine
But what is relevant is the `device_name`. That needs to match the exact `resourceName` of the device which is listed as `permittedHostDevices` in the KubeVirt CR.

I was talking about the `.metadata.name` of the PCIDevice resource earlier. And about the corresponding PCIDeviceClaim's `.metadata.name`.

ibrokethecloud (approved):

Yu-Jack (approved):

PR #193, opened 2026-03-31
fix: use pinned versions and checksum validation

brandboat:
#### Solution:
1. PR ensures pinned versions of artifacts are used and checksum validation is done
2. Align the terraform version to 1.4.6 in both Makefile and Dockerfile

#### Related Issue(s):
https://github.com/rancher/rancher-security/issues/1524

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -4,21 +4,25 @@ ARG DAPPER_HOST_ARCH
 ENV ARCH=$DAPPER_HOST_ARCH
 
 RUN zypper -n rm container-suseconnect && \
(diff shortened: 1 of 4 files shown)


m-ildefons (approved):
LGTM as far as the version pinning is concerned.

However, I think we can make our lives a lot easier by:
1) Using packages from openSUSE repositories where possible.
2) Removing the Dapper installation and just expecting the developer to have it installed on their machine.

Additionally, we may want to replace Terraform with openTofu.

brandboat:
@Mergify backport v1.8

Vicente-Cheng:
@mergifyio backport v1.7

PR #205, opened 2026-05-22
Build System: Replace Dapper

m-ildefons:
Replace Dapper with a basic Make based build system since it's been deprecated.
This PR is a draft until the following requirements are met:

 - [x] `make validate` works and runs the static analysis tools on the code base
 - [x] `make test` works and runs the test suite
 - [x] `make build` works and produces executable binaries in the `bin/` subdirectory
 - [x] `make package` works and produces the expected release artifacts

#### Problem:

Dapper is a central building block of the current build system. But since it is being deprecated, it needs to be replaced - preferably with something that doesn't introduce any additional dependencies.

#### Solution:

Replace Dapper with a basic Make and Docker BuildKit based build system.

#### Related Issue(s):

related-to: https://github.com/harvester/harvester/issues/10623

https://github.com/harvester/harvester/issues/10281

Changes from previous behavior:

1) `make generate` is now part of the `build` target. This makes the build system easier to use, because `make` and/or `make build` will automatically re-generate the documentation.

2) `make` (i.e. the default target) no longer includes running the static analysis tools. The static analysis tools do not accept a dirty Git repository. To make the default target more convenient during development, the validation step is skipped. This makes it possible to re-compile from a dirty Git repository state with `make` or `make build`.

3) The version string is constructed with `git describe --tags --always --dirty`
When checked out at a git tag, this produces the name of the git tag, which means the version numbers for releases stay the same.
However for test builds the version number will now look like e.g.: `v1.8.0-rc1-8-g8b98c01` showing the last tag, the number of commits since that tag, the short commit sha and optionally a `-dirty` indicator when built from a dirty Git tree.
This is much simpler than the previous `version` script.


Dockerfile
```diff
@@ -0,0 +1,152 @@
...
+EOF
+
+# ---- build output ----
+FROM scratch AS build-output
+ARG MK_REPO
+COPY --from=build /go/src/${MK_REPO}/bin/ /bin/
+COPY --from=build /go/src/${MK_REPO}/docs /docs/
+
+# ---- package output ----
+FROM ${MK_PACKAGE_BASE} as package
```
bk201, line 120: since `pakcage`'s Dockerfile is integated to this target, maybe we can condier remove the file https://github.com/harvester/terraform-provider-harvester/blob/master/package/Dockerfile

Dockerfile
```diff
@@ -0,0 +1,152 @@
...
+WORKDIR /go/src/${MK_REPO}
+# to exclude some files, add them in .dockerignore
+COPY . .
+
+# ---- build ----
+FROM base AS build
+ARG MK_REPO
+ARG MK_REPO_ID
+RUN --mount=type=cache,target=/go/pkg/mod,id=harvester-go-mod-${MK_REPO_ID} \
+    --mount=type=cache,target=/go/src/${MK_REPO}/.cache/go-build,id=harvester-go-build-${MK_REPO_ID} \
```
bk201, line 36: nit: personally, I feel it's cleaner to still call a script, but this will work too.

Makefile
```diff
@@ -1,26 +1,90 @@
...
-		exit 1; \
-	fi
-	@@chmod +x .dapper.tmp
-	@./.dapper.tmp -v
-	@mv .dapper.tmp .dapper
-
-$(TARGETS): .dapper
-	./.dapper $@
+ROOT := $(realpath $(dir $(realpath $(firstword $(MAKEFILE_LIST)))))
+comma :=,
```
bk201: comma is not used in the Makefile

Makefile
```diff
@@ -1,26 +1,90 @@
...
+	RESET :=
+else
+	BOLD  := \033[1m
+	CYAN  := \033[36m
+	RESET := \033[0m
+endif
+BANNER = @printf "$(BOLD)$(CYAN)[target: $@]$(RESET)\n"
+
+# Allocate a TTY in dev (for ctrl+c) but not in CI
+MK_DOCKER_RUN_OPTS_TTY := $(if $(CI),,-it)
```
bk201: nit: not used, suggest removed

Dockerfile
m-ildefons: You're right. I missed those. Done.

Dockerfile
m-ildefons: Agreed, this makes the Dockerfile a bit harder to read. But it's a tradeoff.
I put the scripts into heredocs within the Dockerfile because this makes the build system more self-contained.
Since the build system likely (hopefully) won't receive many changes on a regular basis, I'm ok with a bit of ugliness if that means the rest of the directory tree can be cleaner.
Functionally, there is no difference.

Makefile
m-ildefons: Removed.

Makefile
m-ildefons: Removed.

Dockerfile
bk201: I'm OK with the change if the main maintainers of this repo are on the same page. @brandboat

bk201 (approved):

PR #207, opened 2026-06-04
fix(schema): clear all computed-only-illegal attributes in DataSourceSchemaWrap

### LINKED ISSUE #10729: [terraform-provider] Provider fails InternalValidate when a schema field uses DiffSuppressFunc/ValidateDiagFunc
**What happened:**

When a resource schema field uses `DiffSuppressFunc` (or `ValidateDiagFunc`), the corresponding data source — whose schema is produced by `DataSourceSchemaWrap` — fails `provider.InternalValidate` at startup with:

```
DiffSuppressFunc is for suppressing differences between config and state representation. There is no config for computed-only field, nothing to compare.
```

The provider then fails to load.

**Root cause:**

`DataSourceSchemaWrap` makes fields computed-only but does not clear all attributes that are illegal on computed-only fields (`DiffSuppressFunc`, `DiffSuppressOnRefresh`, `ValidateDiagFunc`, `StateFunc`, `InputDefault`).

**Environment:** terraform-provider-harvester (current master).

A fix is ready (clear the complete set of attributes the SDK rejects on computed-only fields).

jniedergang:
Fixes harvester/harvester#10729

### Problem
Data-source fields are made computed-only by `DataSourceSchemaWrap`, but several attributes are illegal on computed-only fields and make `schema.InternalValidate` fail at provider startup (the provider cannot load). This is triggered as soon as a resource field uses `DiffSuppressFunc`/`ValidateDiagFunc`.

### Fix
Clear the complete set the SDK rejects on computed-only fields: `DiffSuppressFunc`, `DiffSuppressOnRefresh`, `ValidateDiagFunc`, `StateFunc`, `InputDefault` (in addition to the previously-cleared `Default`/`DefaultFunc`/`ValidateFunc`/`ConflictsWith`/`AtLeastOneOf`/`ExactlyOneOf`/`RequiredWith`/`MinItems`/`MaxItems`). `DiffSuppressOnRefresh` must be cleared together with `DiffSuppressFunc`, otherwise InternalValidate fails with "cannot set DiffSuppressOnRefresh without DiffSuppressFunc".

### Test plan
- Added a unit test exercising every cleared attribute (including in nested list and set `Elem` schemas), checking the name/namespace inputs stay configurable, and running `InternalValidate` on the wrapped schema (reproduces the startup failure).

### DIFF
--- a/internal/util/schema.go
+++ b/internal/util/schema.go
@@ -82,13 +82,22 @@ func DataSourceSchemaWrap(s map[string]*schema.Schema) map[string]*schema.Schema
 				}
 			}
 		}
+		// Make the field computed-only and strip every attribute the
+		// terraform-plugin-sdk rejects on a computed-only field (see
+		// (*schema.Schema).InternalValidate); otherwise the data source fails
+		// the provider's InternalValidate at startup.
 		v.ForceNew = false
 		v.Computed = true
-		v.Default = nil
-		v.DefaultFunc = nil
 		v.Optional = false
 		v.Required = false
+		v.Default = nil
+		v.DefaultFunc = nil
+		v.InputDefault = ""
+		v.StateFunc = nil
 		v.ValidateFunc = nil
+		v.ValidateDiagFunc = nil
+		v.DiffSuppressFunc = nil
+		v.DiffSuppressOnRefresh = false
 		v.ConflictsWith = nil
 		v.AtLeastOneOf = nil
 		v.ExactlyOneOf = nil
--- a/internal/util/schema_datasource_test.go
+++ b/internal/util/schema_datasource_test.go
@@ -1,0 +1,108 @@
+package util
+
+import (
+	"testing"
+
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/harvester/terraform-provider-harvester/pkg/constants"
+)
+
+func dummyDiffSuppress(_, _, _ string, _ *schema.ResourceData) bool { return false }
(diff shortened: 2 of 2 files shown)


m-ildefons (approved):
LGTM.

Since this wasn't a problem in the past, I assume it only affects some newer Terraform versions.
Would you mind finding out which Terraform versions this affects?
As far as I can tell from the docs, this should be v1.15 or newer, is that right?

jniedergang:
I dug into this and it turns out to be independent of the Terraform CLI version. The check doesn't live in Terraform core but in terraform-plugin-sdk: `Schema.InternalValidate` rejects `DiffSuppressFunc` (and the other attributes cleared here) on computed-only fields. That block predates the SDK split (it's already in terraform 0.12's helper/schema and in plugin-sdk v1.0.0), so every SDK release this provider has ever used contains it. I also went through the Terraform 1.15 changelog and found nothing related.

At runtime Terraform triggers it via `PrepareProviderConfig` -> `Provider.Validate()` -> `InternalValidate()`, which runs on every plan/validate/apply whatever the CLI version.

The reason it never surfaced is that nothing in our resource schemas currently sets `DiffSuppressFunc`/`ValidateDiagFunc`/`StateFunc`, so `DataSourceSchemaWrap` never had anything illegal to copy over. It's latent: as soon as a resource field gains one of those attributes, every data source wrapping that schema makes the provider fail to load (that's what the new unit test reproduces). So no released version is affected, this is just a guard so the next schema change doesn't trip on it.

Refs:
- the check: https://github.com/hashicorp/terraform-plugin-sdk/blob/v2.37.0/helper/schema/schema.go#L1154-L1158
- the runtime path: https://github.com/hashicorp/terraform-plugin-sdk/blob/v2.37.0/helper/schema/provider.go#L309-L323

m-ildefons:
Ah ok, I understand now. Thanks for the explanation. And thanks for providing this fix.

brandboat:
@Mergify backport v1.9

PR #209, opened 2026-06-09
fix: use go.mod

pohanhuang:
#### Problem:
setup-go hard-code the go version, which is not easy to maintain

#### Solution:
use `go-version-file: go.mod` to sync the go version with go.mod


brandboat (approved):
LGTM

PR #210, opened 2026-06-11
Feat: Add SR-IOV GPU device support

m-ildefons:
#### Problem:

The Harvester Terraform provider currently lacks the ability to configure SR-IOV GPU devices, because there is no implementation of corresponding primitives for the respective CRDs of the Harvester PCI device controller.

#### Solution:

Implement Terraform data source and Terraform resources for configuring SR-IOV GPU devices.

#### Related Issue(s):

related-to: harvester/harvester#6424

#### Test plan:

TBD.

Example for enabling SR-IOV on a GPU device:

```
resource "harvester_sriov_gpu_device" "hp-1992-gpu" {
  name = "hp-1992-000026000"
  enabled = true
}
```

Example for enabling and configuring a vGPU Device.
This will cause a corresponding PCI device to show up, which will already be enabled for PCI passthrough.

```
resource "harvester_vgpudevice" "hp-1992-000026042" {
  depends_on = [
    harvester_sriov_gpu_device.hp-1992-gpu
  ]

  name = "hp-1992-000026042"
  node_name = "hp-1992"
  parent_gpu_device_address = "0000:26:00.0"
  enabled = true
  vgpu_type = "NVIDIA L4-2A"
}
```


internal/provider/sriovgpudevice/datasource_sriovgpudevice.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+}
+
+func dataSourceSRIOVGPUDeviceRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	c, err := meta.(*config.Config).K8sClient()
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	name := d.Get(constants.FieldCommonName).(string)
+	pcidevice, err := c.HarvesterDeviceClient.DevicesV1beta1().SRIOVGPUDevices().Get(ctx, name, metav1.GetOptions{})
```
ibrokethecloud: nit, should we rename variable to `sriovGPUDevice`?

internal/provider/sriovgpudevice/schema_sriovgpudevice.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/harvester/terraform-provider-harvester/internal/util"
+	"github.com/harvester/terraform-provider-harvester/pkg/constants"
+)
+
+func Schema() map[string]*schema.Schema {
+	s := map[string]*schema.Schema{
+		constants.FieldSRIOVGPUDeviceEnabled: {
```
ibrokethecloud, line 12: in the current schema the schema is only allowing query of the `SRIOVGPUDevice` against the cluster. should there also be an optional/required value associated with it, similar to what we have with `SRIOVNetworkDevice` https://github.com/harvester/terraform-provider-harvester/blob/master/internal/provider/sriovdevice/schema_sriovnetworkdevice.go#L15 which will be needed to perform actual enable disable operation. 

in a quick test, all i can do is disable the actual gpu device, since `enabled` cannot be passed, it is computed as `false` and device is disabled

```
resource "harvester_sriov_gpu_device" "sriov_gpu_devices" {
  name = "hp-1992-000026000"
}
```

the device cannot be enabled since the field cannot be set by the user

```
│ Error: Value for unconfigurable attribute
│
│   with harvester_sriov_gpu_device.sriov_gpu_devices,
│   on devices.tf line 3, in resource "harvester_sriov_gpu_device" "sriov_gpu_devices":
│    3:   enabled = true
│
│ Can't configure a value for "enabled": its value will be decided automatically based on the result of applying this configuration.
```

ibrokethecloud (commented):
we also need some additional logic to poll and check vf count and names similar to the logic we have sriov network devices.

after making a change to allow `enabled` to be optional, the resource does not report the newly created devices since it can take a few seconds for pcidevices controller to reconcile the newly create vgpudevices

for example

```
 gauravmehta@ubuntu  ~/.../repos/tf-vgpu-test  k get sriovgpudevices
NAME                ADDRESS        NODE NAME   ENABLED   VGPUDEVICES
hp-1992-000026000   0000:26:00.0   hp-1992     true      ["hp-1992-000026004","hp-1992-000026005","hp-1992-000026016","hp-1992-000026017","hp-1992-000026020","hp-1992-000026021","hp-1992-000026022","hp-1992-000026023","hp-1992-000026024","hp-1992-000026025","hp-1992-000026026","hp-1992-000026027","hp-1992-000026006","hp-1992-000026030","hp-1992-000026031","hp-1992-000026032","hp-1992-000026033","hp-1992-000026034","hp-1992-000026035","hp-1992-000026036","hp-1992-000026037","hp-1992-000026040","hp-1992-000026041","hp-1992-000026007","hp-1992-000026042","hp-1992-000026043","hp-1992-000026010","hp-1992-000026011","hp-1992-000026012","hp-1992-000026013","hp-1992-000026014","hp-1992-000026015"]
 gauravmehta@ubuntu  ~/.../repos/tf-vgpu-test  terraform show
# harvester_sriov_gpu_device.sriov_gpu_devices:
resource "harvester_sriov_gpu_device" "sriov_gpu_devices" {
    enabled         = true
    id              = "hp-1992-000026000"
    name            = "hp-1992-000026000"
    vf_addresses    = []
    vf_device_names = []
}
```

m-ildefons:
> we also need some additional logic to poll and check vf count and names similar to the logic we have sriov network devices.

This is a bit problematic, since SR-IOV GPU devices seem to have a fixed number of VFs (depending on manufacturer and model), which isn't known a priori by Terraform. This means when you enable SR-IOV for such a device, eventually some VFs will show up as PCI devices, but to Terraform it's unkonwn how many there will be in the end.
This is different from the SR-IOV network devices, where Terraform knows how many VFs to expect, since the user chooses the number of VFs.

I'm not sure if this is just a design flaw in the PCI device controller, or if this is dictated by the GPU vendors.
As an alternative, I can make the Terraform provider wait until at least one VF shows up as PCI device, or until the number of PCI devices hasn't changes for some time.

ibrokethecloud:
for GPU devices the workflow would be:

* enable device, wait for VF count to be update and just report the VF count and VF devices from status, as they are updated in one go
* disable device, wait for VF count to drop to 0, and then just report the new VF count and VF devices from status as above, except in this case it will be 0 and an empty list.

m-ildefons:
What role does the [VGPUDevice](https://github.com/harvester/pcidevices/blob/bdd80256a8548a4271f6ec4928e2203441a54bc0/pkg/apis/devices.harvesterhci.io/v1beta1/sriovgpudevice.go#L34) CRD play in all of this? There is very little documentation on the API design for this.
Is it necessary to configure a VGPUDevice CRD before the PCI device corresponding to that VF can be passed through to a VM?

internal/provider/sriovgpudevice/resource_sriovgpudevice.go
```diff
@@ -0,0 +1,186 @@
...
+			state = devicesv1.DeviceDisabled
+		}
+		return obj, state, err
+	}
+}
+
+func countPCIDevices(ctx context.Context, client *client.Client, sriovGPUDevice *devicesv1.SRIOVGPUDevice) (int, error) {
+	list, err := client.HarvesterDeviceClient.
+		DevicesV1beta1().
+		PCIDevices().
```
ibrokethecloud, line 167: we are waiting for pcidevices objects to be created, which can take a few mins to show up. would it make sense to use vgpudevices for counting? in my testing it just takes long for reconcile of pcidevices to complete.

internal/provider/vgpudevice/schema_vgpudevice.go
```diff
@@ -0,0 +1,65 @@
...
+			Description: "Configured vGPU type for this vGPU device",
+			Optional:    true,
+		},
+		constants.FieldVGPUDeviceParentGPUDeviceAddress: {
+			Type:        schema.TypeString,
+			Description: "PCI bus address of parent GPU device",
+			Optional:    true,
+		},
+		constants.FieldVGPUDeviceNodeName: {
+			Type:        schema.TypeString,
```
ibrokethecloud, line 30: we do not need to pass nodeName as an argument as the VGPUDevice naming is combination of `nodeName-PCIAddress-Of-VGPU` so it will be unique across the cluster.
Same for `FieldVGPUDeviceParentGPUDeviceAddress`
`FieldVGPUDeviceType` should be mandatory as users need to specify a profile type.

In my test I just specified vgpu resource as follows

```
resource "harvester_vgpudevice" "hp-1992-000026006" {
  name = "hp-1992-000026006"
}
```

end result was vgpudevice on the underlying host was updated

```
apiVersion: devices.harvesterhci.io/v1beta1
kind: VGPUDevice
metadata:
  creationTimestamp: "2026-07-24T04:15:29Z"
  generation: 2
  labels:
    harvesterhci.io/parentSRIOVGPUDevice: hp-1992-000026000
  name: hp-1992-000026006
  resourceVersion: "34350286"
  uid: f40a622c-0194-4d47-9a32-5b422be3a6ba
spec:
  address: "0000:26:00.6"
  enabled: false
  nodeName: ""
  parentGPUDeviceAddress: ""
  vGPUTypeName: ""
status:
  availableTypes:
    NVIDIA L4-1A: "918"
    NVIDIA L4-1B: "908"
    NVIDIA L4-1Q: "910"
    NVIDIA L4-2A: "919"
    NVIDIA L4-2B: "909"
    NVIDIA L4-2Q: "911"
    NVIDIA L4-3A: "920"
    NVIDIA L4-3B: "2166"
    NVIDIA L4-3Q: "912"
    NVIDIA L4-4A: "921"
    NVIDIA L4-4Q: "913"
    NVIDIA L4-6A: "922"
    NVIDIA L4-6Q: "914"
    NVIDIA L4-8A: "923"
    NVIDIA L4-8Q: "915"
    NVIDIA L4-12A: "924"
    NVIDIA L4-12Q: "916"
    NVIDIA L4-24A: "925"
    NVIDIA L4-24Q: "917"
```

the `nodeName`, `parentGPUDeviceAddress` and `vGPUTypeName` name were set as empty strings.

ibrokethecloud (commented):
managing the sriovgpudevice via module currently causes an issue, where when sriovgpudevice is disabled, it causes the node label to be removed from sriovgpudevice.

```
  labels:
    nodename: hp-1992
```

this has a side effect as the pcidevices controller uses this label to identify sriovgpudevices on a node and subsequently apply labels on node to ensure nvidia driver toolkit targets the node.

https://github.com/harvester/pcidevices/blob/master/pkg/controller/nodes/node_controller.go#L229

As a result of this behaviour, when a user attempts to disable a `sriovgpudevice` the label is removed while setting the field `.spec.enabled` to `false`

the remove of label causes the node label to be removed, which in turn causes the nvidia driver to be removed, as a result the `sriovgpudevice` is never competely disabled since the driver pod is no longer present for the remote calls.

internal/provider/sriovgpudevice/resource_sriovgpudevice.go
```diff
@@ -0,0 +1,186 @@
...
+		if !ok || (ok && (vfAddresses.(*schema.Set).Len()) < 1) {
+			return obj, devicesv1.DeviceDisabled, nil
+		}
+
+		numPCIDevicesReady, err := countPCIDevices(ctx, c, obj)
+		if err != nil {
+			return obj, constants.StateCommonError, err
+		}
+
+		stateRaw := d.Get(constants.FieldSRIOVGPUDeviceEnabled).(bool)
```
brandboat, line 154: nit: the naming is a bit weird
```suggestion
		enabled := d.Get(constants.FieldSRIOVGPUDeviceEnabled).(bool)
```

internal/provider/sriovgpudevice/resource_sriovgpudevice.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+			return obj, constants.StateCommonError, err
+		}
+
+		if err = resourceSRIOVGPUDeviceImport(d, obj); err != nil {
+			return obj, constants.StateCommonError, err
+		}
+
+		vfAddresses, ok := d.GetOk(constants.FieldSRIOVGPUDeviceVFAddresses)
+		if !ok || (ok && (vfAddresses.(*schema.Set).Len()) < 1) {
```
brandboat, line 145: can we extract `(vfAddresses.(*schema.Set).Len())` to a new variable?

```
        vfCount := vfAddresses.(*schema.Set).Len()
		if !ok || (ok && vfCount < 1) {
```

internal/provider/sriovgpudevice/resource_sriovgpudevice.go
```diff
@@ -0,0 +1,186 @@
...
+			return obj, devicesv1.DeviceDisabled, nil
+		}
+
+		numPCIDevicesReady, err := countPCIDevices(ctx, c, obj)
+		if err != nil {
+			return obj, constants.StateCommonError, err
+		}
+
+		stateRaw := d.Get(constants.FieldSRIOVGPUDeviceEnabled).(bool)
+		if stateRaw && (vfAddresses.(*schema.Set).Len()) != 0 && numPCIDevicesReady == (vfAddresses.(*schema.Set).Len()) {
```
brandboat, line 155: ditto

internal/provider/sriovgpudevice/schema_sriovgpudevice.go
```diff
@@ -0,0 +1,41 @@
...
+			Type:        schema.TypeBool,
+			Description: "SRIOV enabled",
+			Optional:    true,
```
brandboat, line 15: It's a bit tricky to make this field `optional`,

If user omits enabled, d.Get("enabled") returns false. So this config:

```
resource "harvester_sriov_gpu_device" "gpu0" {
  name = "hp-1992-000026000"
}
```

can unintentionally disable the device. People may think doing this is enabling the gpu_device, but actually it's not. so the optional `enabled` here actually means explicit false, I'm a bit worry about this, what about making this field mandatory?

internal/provider/vgpudevice/schema_vgpudevice.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/harvester/terraform-provider-harvester/pkg/constants"
+)
+
+func Schema() map[string]*schema.Schema {
+	s := map[string]*schema.Schema{
+		// --- read/write properties ---
+		constants.FieldVGPUDeviceEnabled: {
+			Type:        schema.TypeBool,
+			Description: "Enable vGPU for passthrough",
+			Optional:    true,
```
brandboat, line 16: ditto

pkg/importer/resource_vgpudevice_importer.go
```diff
@@ -0,0 +1,29 @@
...
+		constants.FieldVGPUDeviceStatus:                 obj.Status.VGPUStatus,
+		constants.FieldVGPUDeviceUUID:                   obj.Status.UUID,
+		constants.FieldVGPUDeviceConfiguredVGPUType:     obj.Status.ConfiguredVGPUTypeName,
+		constants.FieldVGPUDeviceAvailableTypes:         obj.Status.AvailableTypes,
+	}
+
+	return &StateGetter{
+		ID:           helper.BuildID("", obj.Name),
+		Name:         obj.Name,
+		ResourceType: constants.ResourceTypeSRIOVGPUDevice,
```
brandboat, line 26: It should be `ResourceTypeVGPUDevice`, right?
```suggestion
		ResourceType: constants.ResourceTypeVGPUDevice,
```

internal/provider/sriovgpudevice/resource_sriovgpudevice.go
m-ildefons: When we want to attach a vGPU to a VM, we need the corresponding PCI device to be there.
Therefore we need to wait for the PCI devices to show up. It's not enough to have the vGPU devices.

internal/provider/vgpudevice/schema_vgpudevice.go
m-ildefons: I've asked you to explain the VGPUDevice resource multiple times to no avail.
I still don't understand what each of these fields do exactly and why it is there and what the expected input is.
They are nested under `.spec`, so I would expect, that they should be set by the user as `.spec` should contain the desired state of the system.

Whether they can be left empty or are required is also not clear from reading the PCI device controller code. Shouldn't be there a webhook that ensures that no bogus VGPUDevice CRD can be created?

I also think it's poor engineering to rely on naming conventions to convey information like node name and PCI bus address. It's unreliable at best to marshal this information from the name ad-hoc wherever it's needed.

m-ildefons:
> managing the sriovgpudevice via module currently causes an issue, where when sriovgpudevice is disabled, it causes the node label to be removed from sriovgpudevice.
> 
> ```
>   labels:
>     nodename: hp-1992
> ```
> 
> this has a side effect as the pcidevices controller uses this label to identify sriovgpudevices on a node and subsequently apply labels on node to ensure nvidia driver toolkit targets the node.
> 
> https://github.com/harvester/pcidevices/blob/master/pkg/controller/nodes/node_controller.go#L229
> 
> As a result of this behaviour, when a user attempts to disable a `sriovgpudevice` the label is removed while setting the field `.spec.enabled` to `false`
> 
> the remove of label causes the node label to be removed, which in turn causes the nvidia driver to be removed, as a result the `sriovgpudevice` is never competely disabled since the driver pod is no longer present for the remote calls.

Good grief, the PCI device controller is full of footguns and boobytraps.

Maybe it should be documented which labels are mandatory and what function they fulfill.

Somehow there are also constants, which imply that there should be a different label for this function: https://github.com/harvester/pcidevices/blob/9f5e69ffd97fea14aaa4c0e485b2d894f83c13dc/pkg/apis/devices.harvesterhci.io/v1beta1/sriovgpudevice.go#L70

There is also a `.spec.nodeName` property, which I presume also contains the node name where the GPU hardware is located. Why is this information duplicated between the label and the `.spec` properties? What keeps these in sync and what happens if they get out of sync?

Vicente-Cheng:
Hi @m-ildefons, 
Could we have some Terraform test results that include your changes?
That would make it easy for the reviewer to know the feature works with the implementation.

Thanks!

PR #302, opened 2026-08-06
VM State: Fix VM state for network interfaces

m-ildefons:
Fix VM state for network interfaces with wait_for_lease enabled. If wait_for_lease is enabled for an interface, the VM can not be marked ready before that interface has received an IP address. However, IP addresses may not be assigned immediately as link-local addresses are ignored.
This means that the State() function needs to account for the possibility that an interface doesn't have an IP address at all - not even an empty string assigned.

#### Problem:

If a network interface has `wait_for_lease` enabled, and the IP address isn't assigned quickly enough, the state import accidentally marks the VM as ready.

Marking the VM as ready too early is a problem e.g. when using a remote-exec provisioner, since it relies on being able to establish a network connection.

#### Solution:

Account for network interfaces, which don't have any IP address assigned at all.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/11357

#### Test plan:

```
resource "harvester_virtualmachine" "testvm" {
  provisioner "remote-exec" {
    inline = [
      "echo 'Waiting for cloud-init to complete...'",
      "cloud-init status --wait > /dev/null",
      "echo 'Completed cloud-init!'",
    ]

    connection {
      type        = "ssh"
      host        = self.network_interface[0].ip_address
      user        = var.ssh_user
      certificate = var.ssh_pub_key[0]
      agent       = true
      script_path = "/tmp/user-data-check.sh"
    }
  }

  network_interface {
    name         = "nic-1"
    network_name = data.harvester_network.vm_network.id
    wait_for_lease = true
  }

  cloudinit  {
    user_data_secret_name = harvester_cloudinit_secret.bootstrap.name
    network_data_secret_name = harvester_cloudinit_secret.bootstrap.name
  }

  timeouts {
    create = "10m"
  }
}
```

https://suse.slack.com/archives/C02CN3TT61M/p1785948849015539

### DIFF
--- a/pkg/importer/resource_cloudinitsecret_importer_test.go
+++ b/pkg/importer/resource_cloudinitsecret_importer_test.go
@@ -1,0 +1,43 @@
+package importer
+
+import (
+	"testing"
+
+	assert "github.com/stretchr/testify/assert"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+
+	"github.com/harvester/terraform-provider-harvester/pkg/constants"
+)
+
+func TestStateRunningNetworkInterfaceNoIP(t *testing.T) {
+	// Ensure that of `wait_for_lease` is set on a network interface, the state isn't reported as
+	// `ready` as long as there is no IP address
+	oldInstanceUID := "oldUid"
+	networkInterfaces := []map[string]any{
+		{
+			constants.FieldNetworkInterfaceWaitForLease: true,
+		},
+	}
+	vmi := kubevirtv1.VirtualMachineInstance{
+		ObjectMeta: metav1.ObjectMeta{
+			UID: "newUid",
+		},
+		Status: kubevirtv1.VirtualMachineInstanceStatus{
+			Phase: "Running",
+		},
+	}
+
+	vmImporter := NewVMImporter(nil, &vmi)
+	state := vmImporter.State(networkInterfaces, oldInstanceUID)
+	assert.Equal(t, state, constants.StateVirtualMachineRunning, "IP address not set results in state running")
+
+	networkInterfaces[0][constants.FieldNetworkInterfaceIPAddress] = ""
+	state = vmImporter.State(networkInterfaces, oldInstanceUID)
+	assert.Equal(t, state, constants.StateVirtualMachineRunning, "IP address set to empty string results in state running")
+
+	networkInterfaces[0][constants.FieldNetworkInterfaceIPAddress] = "123.123.123.123"
+	state = vmImporter.State(networkInterfaces, oldInstanceUID)
+	assert.Equal(t, state, constants.StateCommonReady, "IP address set to non-empty string results in state ready")
+}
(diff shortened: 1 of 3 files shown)


pkg/importer/resource_virtualmachine_importer.go
```diff
@@ -357,8 +357,11 @@ func (v *VMImporter) State(networkInterfaces []map[string]interface{}, oldInstan
 		if string(v.VirtualMachineInstance.UID) == oldInstanceUID {
 			return constants.StateVirtualMachineRunning
 		}
+
 		for _, networkInterface := range networkInterfaces {
-			if networkInterface[constants.FieldNetworkInterfaceWaitForLease].(bool) && networkInterface[constants.FieldNetworkInterfaceIPAddress] == "" {
+			wait_for_lease := networkInterface[constants.FieldNetworkInterfaceWaitForLease].(bool)
```
w13915984028, line 362: Is there a general timeout for this wait ?

There are a couple of cases for interface to show no IPs:
Fail to get IP from DHCP server (timeout, no ip, no server ...)
guest-agent is missing/not working, IP is not reported
...

From Harvester UI, we can observe a clear time gap between the VM shows ready and the IP pops up.

m-ildefons: Yes. The maximum timeout is determined by the context for the create action. By default it's two minute, but can be adjusted by the user with a `timeout {}` block in the resource.
If no IP shows up ever, it will eventually reach that timeout and fail resource creation.
If the user wants to have interfaces without IP addresses, they should set `wait_for_lease = false`.

There are many different interpretations of what "ready" means for a VM. And there are likewise many different indicators we can use to determine readiness.
I'm not sure what indicators the Harvester UI goes by, but the behavior for `wait_for_lease` is that the resource creation will only finish when an IP address has been obtained (presumably via a DHCP lease). This makes sense in the context of the Terraform provider because of things like the remote-exec provisioner, which require that the network for the VM is configured.

pkg/importer/resource_cloudinitsecret_importer_test.go
```diff
@@ -0,0 +1,43 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+
+	"github.com/harvester/terraform-provider-harvester/pkg/constants"
+)
+
+func TestStateRunningNetworkInterfaceNoIP(t *testing.T) {
+	// Ensure that of `wait_for_lease` is set on a network interface, the state isn't reported as
+	// `ready` as long as there is no IP address
+	oldInstanceUID := "oldUid"
```
w13915984028, line 17: how was `oldInstanceUID` really used in this function? the test code does not over the direct return path

https://github.com/harvester/terraform-provider-harvester/pull/302/changes#diff-3b4401b6126ec3a8f73bce2129fef5b6959f6a1fb178b384af2a905593a26cd3R357

w13915984028 (approved):
LGTM, thanks.

pkg/importer/resource_cloudinitsecret_importer_test.go
m-ildefons: I have no idea what `oldInstanceUID` is supposed to do.

This test case intentionally doesn't cover the direct return path, or any other return path of that function, because I want this PR to stay in scope and only focus on the defect.

w13915984028: no worries, just skip it for now, it is out of the scope of this PR.

brandboat:
@Mergify backport v1.9

PR #320, opened 2026-08-28
Dependencies: Replace yaml library

m-ildefons:
Replace `gopkg.in/yaml.v2` with `go.yaml.in/yaml/v3`

#### Problem:

Use of unmaintained YAML library.

Unmarshalling data using this YAML library targets a looser than necessary data type.

#### Solution:

Replace `gopkg.in/yaml.v2` with `go.yaml.in/yaml/v3`.
Replace `map[interface{}]interface{}` with `map[string]any` as datatype for the destination of `yaml.Unmarshal()`. This is required by the v3 version of the YAML library `go.yaml.in/yaml`

#### Related Issue(s):

related-to: harvester/harvester#11234

#### Test plan:

No functional changes expected.

A number of new unit tests have been added to make sure the map constructed during unmarshalling with the new YAML library is sane and any modified code behaves as expected.


internal/provider/virtualmachine/resource_virtualmachine_validator.go
```diff
@@ -88,27 +88,27 @@ func checkKeyPairsInUserData(userdataContent []byte, keyPairs []*harvsterv1.KeyP
 		}
 	}
 	if len(missingKeyPairs) > 0 {
-		return fmt.Errorf(`missing ssh public keys in cloud-int userdata ssh_authorized_keys section, ssh_keys: [%s].
-Either remove unused ssh keys from "ssh_keys" or add ssh public keys to cloud-int userdata ssh_authorized_keys section. `, strings.Join(missingKeyPairs, ","))
+		return fmt.Errorf(`missing ssh public keys in cloud-int userdata ssh_authorized_keys section: %v, ssh_keys: [%s].
+Either remove unused ssh keys from "ssh_keys" or add ssh public keys to cloud-int userdata ssh_authorized_keys section. `, userData, strings.Join(missingKeyPairs, ","))
```
brandboat: I'm not sure if it's ok to print userData in the error message, userData can container some sensitive info like password, token...

brandboat (dismissed):
Overall LGTM.

internal/provider/virtualmachine/resource_virtualmachine_validator.go
m-ildefons: Good catch. This was left over from debugging earlier.

brandboat (approved):
Thanks!

brandboat:
@Mergify backport v1.9

PR #334, opened 2026-09-11
deps: remove k8s.io/utils dependency

m-ildefons:
Since Golang v1.26 the builtin `new()` accepts value expressions. This is functionally equivalent to `ptr.To()` from the k8s.io/utils library. Therefore the dependency on the k8s.io/utils library can be removed as the only use of this library in the code base was `ptr.To()`.

#### Problem:

Unnecessary dependency on external library.

#### Solution:

Replace uses of `ptr.To()` with Golang builtin `new()`. This makes the import of the library superfluous.

#### Related Issue(s):

N/A

#### Test plan:

N/A. No functional changes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,7 +42,6 @@ require (
 	k8s.io/api v0.34.1
 	k8s.io/apimachinery v0.34.1
 	k8s.io/client-go v12.0.0+incompatible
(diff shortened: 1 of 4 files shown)


brandboat (approved):
Thanks for the cleanup!

brandboat:
@Mergify backport v1.9

